PR #125, opened 2014-12-28
Fix Xcode 6 LLVM compiler error

stevenp:
See: http://stackoverflow.com/questions/24922913/too-many-arguments-to-function-call-expected-0-have-3

By default, Xcode now enforces strict checking of objc_msgSend calls. When trying to integrate PromiseKit as a dependency for a new pod I'm creating, I was unable to compile because of this issue.

### DIFF
--- a/objc/UIViewController+PromiseKit.m
+++ b/objc/UIViewController+PromiseKit.m
@@ -116,7 +116,7 @@ static void classOverridingSelector(const char* newClassPrefix, id target, SEL o
         struct objc_super objcSuper;
         objcSuper.receiver = self;
         objcSuper.super_class = zuper;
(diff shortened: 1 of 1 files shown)


dmiotti:
Hi, any plan when this will be available through cocoapods ? :)

krzysztofzablocki:
@mxcl +1 on this, annoying to modify pod source each time I update just for this right now

mxcl:
Sorry, I've been offline. Reviewing now.

mxcl:
I'm tempted to remove the promiseSegue stuff altogether. It's way too hacky and potentially dangerous and I'm not convinced anyone uses it.

It was a PR a while ago.

Issue #115, opened 2014-11-30
When adding swift/PromiseKit.xcodeproj, a linker error occurs if you never `import AssetsLibrary`


dimroc:
Given a project where you never `import AssetsLibrary`, i.e.:

```swift
import Foundation
//import AssetsLibrary
import PromiseKit
```

Running your project will result in the following:

```bash
dyld: Library not loaded: @rpath/libswiftAssetsLibrary.dylib
  Referenced from: /Users/dimroc/Library/Developer/Xcode/DerivedData/ManhattanForum-ekevinzlbwehjbdjpibmdbxqdqtm/Build/Products/Debug-iphonesimulator/PromiseKit.framework/PromiseKit
  Reason: image not found
```

This might be due to the following lines of code in `PromiseKit/swift/Sources/UIViewController+Promise.swift`:

```swift
4: import AssetsLibrary.ALAssetsLibrary
```

Simply uncommenting the `import AssetsLibrary` in the main application, cleaning, and building fixes the problem.

I'm not sure there is an explicit fix for this, especially with the inevitable cocoapod support for swift. I figured I would document it here in case other people run into the issue. Is there a stackoverflow tag I should use instead?

mxcl:
The new `PromiseKit/Swift` pod imports AssetsLibrary. However I'd rather load it at runtime on demand. Since it is used only to provide the raw NSData for the image, if the user asks for that version of the Promise, so mostly it is a waste to load. So I'll leave this ticket open with that goal in mind.

hariseldon78:
i have the same problem with the CoreAudio library: 
dyld: Library not loaded: @rpath/libswiftCoreAudio.dylib
  Referenced from: /Users/robertoprevidi/Developer/moving-circles/Build/Products/Debug-iphonesimulator/PromiseKit.framework/PromiseKit
  Reason: image not found

mxcl:
Well, this isn't to do with PromiseKit then. We don’t do anything with CoreAudio. You should be able to fix it however by making your *app* target link to CoreAudio.

PR #130, opened 2015-01-12
Adding Swift subspec for building PromiseKit as a Swift framework

mowens:
Adding new "Swift" subspec so developers can install PromiseKit as a swift framework (requires 0.36 beta build of CocoaPods)

```ruby
pod 'PromiseKit/Swift', '~> 1.2'
```

### DIFF
--- a/PromiseKit.podspec
+++ b/PromiseKit.podspec
@@ -171,6 +171,10 @@ Pod::Spec.new do |s|
     ss.dependency 'PromiseKit/QuartzCore'
   end
 
+  s.subspec 'Swift' do |ss|
+    ss.source_files = 'swift/Sources/**/*.{swift,h,m}'
+  end
+
 #### deprecated
 
   s.subspec 'SKProductsRequest' do |ss|


ldiqual:
Would it be possible to also have subspecs for `Promise` and the different extensions? Somthing like `PromiseKit/Swift/Promise` or `PromiseKit/Swift/NSURLConnection`.
I'm facing this issue: https://github.com/mxcl/PromiseKit/issues/115 and my solution with Obj-C was to only include `PromiseKit/Promise` so extra headers wouldn't get included. With this PR, there's no way to be finer with what I want in my project.

mxcl:
I don’t think you can have sub-subspecs, but I'd like to offer the same modularity options for both.

ldiqual:
Oh, apparently you can have nested subspecs: http://guides.cocoapods.org/syntax/podspec.html#subspec (end of the page). It'd be awesome to have that!

mxcl:
@ldiqual good find. I'll merge this but aim to add that with an update.

mxcl:
I started adding nesting subspecs, but only as I've needed them. PR welcome! Or I will get round to eventually.

PR #138, opened 2015-02-01
+whenResolved:

tikitu:
Add a variant of +when: which does not short-circuit on failure of one of the promises in the array.

It demands an NSArray as input rather than accepting a single promise or dictionary as +when: does. Always resolves to two values: an NSArray of resolved values and an NSArray (or nil) of errors. There is no way to map the results and errors back to their originating promises.

### DIFF
--- a/objc/PMKPromise+Join.m
+++ b/objc/PMKPromise+Join.m
@@ -1,12 +1,12 @@
 #import <Foundation/NSArray.h>
 #import <Foundation/NSDictionary.h>
 #import "PromiseKit/Promise+Until.h"
-#import "PromiseKit/Promise+WhenResolved.h"
+#import "PromiseKit/Promise+Join.h"
 
 
-@implementation PMKPromise (WhenResolved)
+@implementation PMKPromise (Join)
 
-+ (PMKPromise *)whenResolved:(NSArray*)promises {
++ (PMKPromise *)join:(NSArray*)promises {
     __block NSMutableArray *mutablePromises = [promises mutableCopy];
     __block NSMutableArray *collectedErrors = [NSMutableArray new];
 
--- a/objc/PMKPromise+WhenResolved.m
+++ b/objc/PMKPromise+WhenResolved.m
@@ -1,0 +1,22 @@
+#import <Foundation/NSArray.h>
+#import <Foundation/NSDictionary.h>
+#import "PromiseKit/Promise+Until.h"
+#import "PromiseKit/Promise+WhenResolved.h"
+
+
+@implementation PMKPromise (WhenResolved)
+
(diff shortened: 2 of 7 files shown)


mxcl:
I'm cool with this, but I'd prefer a different name. I'll think on it. I'm not sure there is any precedent. Eg. Promises with ES6 have a `race`, function which is all we don't have, but it is not the same as your contribution.

tikitu:
If you find a precedent that suggests a different API (or have any better ideas) I'm happy to work on it further. (Particularly unsure about not being able to map results back to their originating promises, but I think making that possible would complicate things quite a lot.)

mxcl:
Maybe `wait`?

tikitu:
I'm not totally sold on `wait`, it sounds a bit like it should be blocking. (Your project = your call, just my 2c.) Some alternatives:

* `after`
* `afterAll`
* `whenFinally` (by analogy with `.finally()` which collapses the reject/fulfill distinction)

If you make a decision I'll rebase the PR and make the changes.

mxcl:
How about `join`?

tikitu:
I like. Will do renaming and rebasing; if you come up with something better in the meantime it's no trouble to redo.

mxcl:
Many thanks, an excellent contribution.

tikitu:
Huh, I was still trying to rebase at my end (with not much success, I must say). Thanks!

mxcl:
np, I rewrote the tests to use XCTestExpectation, so the rebase would have been a PITA. But also I didn't realize you weren't finished, so my bad! All worked out though. The merge is fine IMO.

mxcl:
Can I ask for an example use-case for join?

I've seen people using it instead of `when` and thus breaking their error handling, which disturbs me, if I understood more about what the requirements were that made this method get added, I could document it better.

tikitu:
Our original use case was uploading a bunch of items that were created offline to an API server when next online. We want to track failures, but just because one item fails doesn't mean we should abort all the other uploads.

mxcl:
Well, the other uploads won't abort. PromiseKit cannot cancel anything. That's up to the developer should they want that.

However when does not let you know which succeed. So indeed.

tikitu:
Two other considerations from that use case:
* if we short-circuit on first failure, as you say nothing gets explicitly cancelled, but we do start having to worry about retaining references long enough to make sure the other requests don't get garbage-collected;
* with short-circuiting it is not easy to defer an action until all the results are in (e.g. reporting "3 uploads succeeded and 4 failed").

mxcl:
I added some more documentation.

With hindsight probably join should have rejected the promise still if an error occurs, but still it should wait until all promises are resolved first.

PR #143, opened 2015-02-08
add  POST with json type equals to [String:AnyObject]

pgherveou:
Otherwise everything is encoded as a string and we can't call the api like this 
```swift
NSURLConnection.POST(url, JSON: ["foo": 1, "bar": "..."])
```


mxcl:
Yes, but AnyObject isn't valid JSON. You can have strings, numbers, bools, arrays or dictionaries of the same. I was kind of waiting on some Pod that does JSON better and would then use that.

Still, practically, this is what you need in the interim so I guess it should be committed.

pgherveou:
That would be great, unless you see another work around?

PR #152, opened 2015-02-25
update project framework search paths

pgherveou:
following up my issues trying to compile the oroject on my own machine, i think these project settings can be removed


mxcl:
I think we just need to make the path not hardcoded to my machine.

Generally, the project needs OMGHTTPURLRQ and when I work on it locally I don't want to have to always remember to tweak the framework paths.

Issue #133, opened 2015-01-20
Convenience then for running on background thread


dbachrach:
I find myself frequently doing:

```objc
[object fetch].thenOn(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^(id response) {
    // do something
}
```

What's nice about `then` is that is succinct and clear-- it always runs on the main thread. I think a large majority of promise chains will need to `then` onto the main thread or `then` onto a generic background thread. Just like there is a succinct and clear API to `then` on to the main queue, what are your thoughts towards a succinct and clear API to `then` onto a background queue.

```objc
- (PMKPromise *(^)(id))thenBackground {
    return ^(id block){
        return self.thenOn(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), block);
    };
}
```

mxcl:
Agreed. I've almost written something like this many times, I just couldn't decide on a good name.

`thenOnBackground`, `thenOnBG`, etc. something.

mxcl:
Even though it is dirty I often do:

```objc
[object fetch].thenOn(dispatch_get_global_queue(0, 0), ^(id response) {
}
```

dbachrach:
Yeah understood. I think `thenOnBackground` is more objc-like and still pretty terse. I think this is a useful standard to define in the library. Happy to make a PR.

mxcl:
If we follow Apple’s lead we have: https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSObject_Class/index.html#//apple_ref/occ/instm/NSObject/performSelectorInBackground:withObject:

That is: `thenInBackground`.

I guess that is best. But yes, I find it longer in letters than I'd like. Which is why it doesn't exist yet. I guess I was waiting for some further inspiration for a short word like `then` that implied background.

dbachrach:
Good point. In over On. A quick browse of the thesaurus provides:

```objc
[obj fetch].later(^(id response) {
    // in background
});
```

Which is interesting, but vague. I'll put a PR together for `thenInBackground`

PR #154, opened 2015-02-27
thenInBackground for Swift

sdolidze:
Added thenInBackground for Swift (which is already implemented in Objective-C). See #133.


mxcl:
Thanks.

PR #168, opened 2015-03-31
Make LocationManager class internally visible, so that it works as CLLocationManagerDelegate

tp:
With the latest Xcode update `locationManager(manager: CLLocationManager!, didUpdateLocations locations: [AnyObject]!)` got never called in my code, until I randomly changed the class visibility from private to internal.

I think this has worked before. Sadly I was not yet able to pin down a reason why this change is necessary. (or maybe I am doing something wrong and saw the result for un unrelated reason?)

### DIFF
--- a/Swift Sources/CLLocationManager+Promise.swift
+++ b/Swift Sources/CLLocationManager+Promise.swift
@@ -1,6 +1,6 @@
 import CoreLocation.CLLocationManager
 
-private class LocationManager: CLLocationManager, CLLocationManagerDelegate {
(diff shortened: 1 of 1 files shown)


mxcl:
I reported this as a bug to the Swift team, it should get fixed without code changes on our end. So I guess we'll wait for the next beta.

mxcl:
Strangely this seemed to be the only case where `@objc` needed to be specified. And we have a lot of equivalent delegates.

PR #172, opened 2015-04-05
port non-2xx response handling over from objc

allen-zeng:
Currently ```NSURLConnection``` in Swift lib fulfills server non 200 responses.

This change ports the Objective-C functionality to reject non 200 responses over. Starting from [this line](https://github.com/mxcl/PromiseKit/blob/master/objc/NSURLConnection+PromiseKit.m#L86)

### DIFF
--- a/Swift Sources/NSURLConnection+Promise.swift
+++ b/Swift Sources/NSURLConnection+Promise.swift
@@ -23,11 +23,11 @@ private func fetch<T>(var request: NSURLRequest, body: ((T) -> Void, (NSError) -
     }
 
     return Promise<T> { (fulfiller, rejunker) in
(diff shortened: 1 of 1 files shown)


mxcl:
Thanks, I wonder why I didn't do this already.

FFR, the diff on line 44 shouldn’t have made it into the PR.

allen-zeng:
No thank _you_ for all the awesome work! :)

Interesting point on line 44, is it because that it isn't related to the pull request or is it because there is a specific coding style that you'd like to see?

mxcl:
The diff just changes formatting and thus isn't related to the content of your PR.

If you wanted to change the formatting it should be a separate PR.

Consequently the diff for this commit contains unrelated modifications and impedes the legibility of the actual fix.

allen-zeng:
:+1:

PR #173, opened 2015-04-14
Support MFMessageComposeViewController in UIViewController+Promise

nathanhosselton:
Added support for promising instances of MFMessageComposeViewController and handling delegation similar to the existing support for MFMailComposeViewController. The implementations were very similar so this was a straightforward addition.

In testing the Swift counterpart I ran into issues with then'ing the result due to an attempted cast from the corresponding result type (MFMailComposeResult, MessageComposeResult) to Int. I changed the type parameter for each method's returned promise to the explicit result type to fix this issue. Due to my inexperience with Swift, I'm unsure if this is the ideal solution, but it did begin working properly after the change.

### DIFF
--- a/Swift Sources/UIViewController+Promise.swift
+++ b/Swift Sources/UIViewController+Promise.swift
@@ -84,12 +84,12 @@ extension UIViewController {
         return promiseViewController(vc, animated: animated, completion: completion)
     }
 
-    public func promiseViewController(vc: MFMailComposeViewController, animated: Bool = false, completion:(Void)->() = {}) -> Promise<Any> {
+    public func promiseViewController(vc: MFMailComposeViewController, animated: Bool = false, completion:(Void)->() = {}) -> Promise<MFMailComposeResult> {
         vc.mailComposeDelegate = MFComposeViewControllerProxy()
         return promiseViewController(vc as UIViewController, animated: animated, completion: completion)
     }
 
-    public func promiseViewController(vc: MFMessageComposeViewController, animated: Bool = false, completion:(Void)->() = {}) -> Promise<Any> {
+    public func promiseViewController(vc: MFMessageComposeViewController, animated: Bool = false, completion:(Void)->() = {}) -> Promise<MessageComposeResult> {
         vc.messageComposeDelegate = MFComposeViewControllerProxy()
         return promiseViewController(vc as UIViewController, animated: animated, completion: completion)
     }
--- a/Swift Sources/UIViewController+Promise.swift
+++ b/Swift Sources/UIViewController+Promise.swift
@@ -23,7 +23,7 @@ class MFComposeViewControllerProxy: NSObject, MFMailComposeViewControllerDelegat
 
     func messageComposeViewController(controller: MFMessageComposeViewController!, didFinishWithResult result: MessageComposeResult) {
         if result.value == MessageComposeResultFailed.value {
-            controller.reject(NSError(domain: PMKErrorDomain, code: 4, userInfo: [NSLocalizedDescriptionKey: "The user’s attempt to save or send the message was unsuccessful."]))
+            controller.reject(NSError(domain: PMKErrorDomain, code: PMKOperationFailed, userInfo: [NSLocalizedDescriptionKey: "The user’s attempt to save or send the message was unsuccessful."]))
         } else {
             controller.fulfill(result)
         }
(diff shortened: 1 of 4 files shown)


objc/UIViewController+PromiseKit.m
```diff
@@ -6,11 +6,13 @@
 #import <UIKit/UINavigationController.h>
 #import <UIKit/UIImagePickerController.h>
 #import "UIViewController+PromiseKit.h"
+@import MessageUI.MFMessageComposeViewController;
+@import MessageUI.MFMailComposeViewController;
```
mxcl: Can't @import as this imposes framework requirements on our end users. Since objc is dynamic, we don’t need to, hence we were using `NSClassFromString`.

Swift Sources/UIViewController+Promise.swift
```diff
@@ -19,6 +20,15 @@ class MFMailComposeViewControllerProxy: NSObject, MFMailComposeViewControllerDel
         }
         PMKRelease(self)
     }
+
+    func messageComposeViewController(controller: MFMessageComposeViewController!, didFinishWithResult result: MessageComposeResult) {
+        if result.value == MessageComposeResultFailed.value {
+            controller.reject(NSError(domain: PMKErrorDomain, code: 4, userInfo: [NSLocalizedDescriptionKey: "The user’s attempt to save or send the message was unsuccessful."]))
```
mxcl: 4 is `PMKAccessDeniedError`, so we should use the symbol and not a magic number. Still it seems the wrong code. We may need a new code. It is bizarre that this part of MessageUI doesn't provide an `NSError`! I vote for `PMKOperationFailed` which is as generic sounding as Apple should be embarrassed.

objc/UIViewController+PromiseKit.m
nathanhosselton: Ah, makes sense. I wasn't going to import originally but since the MessageCompose delegate method doesn't provide an error I needed to check the result value directly which left me with the option of either importing or assuming the int value of the Failed result.

Swift Sources/UIViewController+Promise.swift
mxcl: The description should be: `The user’s attempt to save or send the message was unsuccessful.` as that is what Apple says in the docs.

nathanhosselton: Ok yes I wasn't sure what error code to use here so I used 4 because I didn't see it used for anything in the swift implementation.

objc/UIViewController+PromiseKit.m
```diff
@@ -145,6 +156,16 @@ - (void)mailComposeController:(id)controller didFinishWithResult:(int)result err
 
     PMKRelease(self);
 }
+
+- (void)messageComposeViewController:(id)controller didFinishWithResult:(MessageComposeResult)result {
+    if (result == MessageComposeResultFailed)
+        [controller reject:[NSError errorWithDomain:PMKErrorDomain code:PMKUnknownError userInfo:@{NSLocalizedDescriptionKey: @"The user’s attempt to save or send the message was unsuccessful."}]];
```
mxcl: We should be consistent with code and message as the Swift variant.

Swift Sources/UIViewController+Promise.swift
mxcl: I guess there isn't a 4 for Swift PMK? Still we should fix it with a new code for both.

mxcl:
Good. Push the fixes and I'll merge.

objc/UIViewController+PromiseKit.m
mxcl: Yes, that does suck. Well it's ok to hard code the int values, they won't change. Apple won't risk it. but it is sucky.

For PMK2 I am changing this sort of stuff so we can @import, by making the end user compile the categories into their own end binaries. I have an easy way to facilitate that so it won't be a PITA.

mxcl:
Shall I make the changes?

nathanhosselton:
Sorry, I'll get to it by this weekend at the latest as long as that's ok. I've just been super busy this week.

mxcl:
There’s no rush, I just offer personal services to friends.

Swift Sources/UIViewController+Promise.swift
nathanhosselton: Alright so what are we doing with the error codes? Right now they don't match at all between Swift and ObjC. Are we going to try to match them up? Or just add a PMKOperationFailed code to both and be done?

mxcl: Add a new one to both I reckon. Make them the same int value.

nathanhosselton:
Changes are in. Let me know if anything further is needed.

PR #178, opened 2015-04-26
fix navigation controller presenting promise

allen-zeng:
Right now calling ```promiseViewController``` with a ```UINavigationController``` just presents its root view controller modally, which just ends up crashing with an ```NSInvalidArgumentException``` that says
>Application tried to present modally an active controller

This change binds the promise to the root view controller and modally presents the navigation controller instead.


mxcl:
Pushing 1.5.2

Issue #181, opened 2015-04-28
Swift thread safety?


xenadu:
While looking at `race()`, I got curious why `thenUnleashZalgo()` doesn't protect mutation of `Handlers` with `dispatch_barrier_sync`. It appears that any use of `thenUnleashZalgo()` has the potential to corrupt the `handlers` array. 

If one queue is chaining the promise at the same time the main queue is calling `thenUnleashZalgo()` then the potential for data races or corruption is present.

Worse, if a background queue is fulfilling it may either observe the handlers array prior to mutation causing the handler to never be called, or it may observe the closure's presence in the array before the closure itself has been fully constructed (regardless of what order those statements happen in the Swift standard library, the compiler and the CPU are free to re-order the writes and publish the change to the array before construction of the closure's environment is captured and fully constructed)

Am I missing something obvious here?

edit: removed mistaken view of how the normal chaining works; I see how it uses the barrier inside `init` to protect everything. I think `thenUnleashZalgo` is the only actual question.

mxcl:
Yes this is an omission.

Is there a good way to test this sort of thing? We'd need to get into the mechanism and cause the kinds of race conditions that would cause corruption.

PR #182, opened 2015-04-28
Tests for data races

xenadu:
A somewhat reliable test for #181. When Zalgo isn't properly protected, the test crashes on every iteration I ran (with memory corruption). When Zalgo is properly mollified the test passes. 

Also included a version for `then()` just to cover the situation where a data race is inadvertently introduced later.

### DIFF
--- a/Specs/Promise.test.swift
+++ b/Specs/Promise.test.swift
@@ -25,6 +25,27 @@ class TestPromise: XCTestCase {
         }
     }
 
+    func stressDataRace<T: Equatable>(e1: XCTestExpectation, iterations: Int = 1000, stressFactor: Int = 10, stressFunction: (Promise<T>) -> Void, fulfill f: () -> T) {
+        let group = dispatch_group_create()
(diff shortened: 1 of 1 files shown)


mxcl:
Thanks, but I can't pull unless it is 100% reliable or Travis will be emailing me with false failure reports.

xenadu:
@mxcl I would never open a PR with a test that has any chance of generating spurious test failures. 

The unreliability of this test is that it can *in theory* pass when it should have failed. Never the other way around... If there is a data race present there is a small chance that the data race occurs in just such a perfect way as to allow the test to pass. I tried to set the iteration and stress counts high enough to make sure that doesn't happen, but it depends on how the system allocates dispatch queues, the number of processors, and other factors outside of our control.

I don't think proving the absence of a race in this context is a computable problem.

mxcl:
aah yes, I should have realized that. My bad.

mxcl:
Well in theory we could mock and stub so we can force the various race conditions, and then prove they don't happen, but I've never tried to do such a thing before. Maybe it's not practical.

PR #190, opened 2015-05-12
PromiseKit 2.0

### LINKED ISSUE #13: Reconsider catching exceptions
Why does PromiseKit propagate exceptions (instances of the class `NSException`) to clients? Wouldn't it be more appropriate to follow the Cocoa convention of using errors (instances of the class `NSError`) exclusively to indicate failure?

Since clients rescue from erros instead of catch exceptions, I suggest to rename `-[Promise catch]` to `-[Promise rescue]` or `-[Promise error]`. If clients write code that throws exceptions, it is the clients fault that should be handled during development. Error handling at runtime should not occur by catching exceptions.

The [Error Handling Programming Guide - A Note on Errors and Exceptions](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-SW6) states the following:

> It is important to keep in mind the difference between error objects and exception objects in Cocoa and Cocoa Touch, and when to use one or the other in your code. They serve different purposes and should not be confused.

> Exceptions (represented by NSException objects) are for programming errors, such as an array index that is out of bounds or an invalid method argument. User-level errors (represented by NSError objects) are for runtime errors, such as when a file cannot be found or a string in a certain encoding cannot be read. Conditions giving rise to exceptions are due to programming errors; you should deal with these errors before you ship a product. Runtime errors can always occur, and you should communicate these (via NSError objects) to the user in as much detail as is required.

> Although exceptions should ideally be taken care of before deployment, a shipped application can still experience exceptions as a result of some truly exceptional situation such as “out of memory” or “boot volume not available.” It is best to allow the highest level of the application—the global application object itself—to deal with these situations."

Let me know what you think.

### LINKED ISSUE #18: Canceling a promise
Consider an asynchronous HTTP GET method request that returns a promise:

```
- (Promise*)GET:(NSString*)path;
```

Users of this API may want to cancel the underlying asynchronous work, the HTTP request. Currently PromiseKit does not allow cancelable promises.

I've been looking into javascript promises and cancelation. There is some discussion in the A+ spec https://github.com/promises-aplus/cancellation-spec/issues

Any thoughts on cancelation?

### LINKED ISSUE #56: Promise<Void> catch->Void is ambiguous
I thought that swift incorrectly define type of a closure, but casting doesn't help. 
Example from README.md doesn't work for me as well.
Is it bug, or i'm doing something wrong?

screenshot

```swift
let promise: Promise<Int> = Promise({ success, failure in
    success(10)
})
        
promise.then { number in
    self.remainingTimeLabel.text = "int is \(number)"
}.catch { error in
    self.remainingTimeLabel.text = "erorr is \(error)"
}
```

### LINKED ISSUE #165: Use PMKPromise with Swift
I have had some troubles using PMKPromise in Swift. We have a big project written in objective-C that uses PromiseKit (PMKPromise) and we would like to start using Swift. 

In more detail, the problem is that when calling ```then``` with a Swift closure as a parameter, the compiler complaints about the types of the parameter.

mxcl:
Fixes #165
Fixes #56
Closes #18
Fixes #13


mxcl:
Here’s what will be the accompanying blog post: http://promisekit.org/PromiseKit-2.0-Released/

djtarazona:
:+1:

PR #195, opened 2015-05-24
Decode response data only when data != nil

newlix:


mxcl:
Thank you. I'll see if I can create a test case for this.

mxcl:
I'm pushing PromiseKit 2.0.1 as we speak.

PR #211, opened 2015-06-27
Fix when() behavior for empty array case

jflinter:
Currently, calling `when` on an empty array returns a promise that will never be resolved. This modifies it to instead return an immediately-fulfilled promise.


mxcl:
Thanks. I am upset this snuck through. The object version has this test.

mxcl:
Pushed 2.0.6

jflinter:
Thanks!

PR #214, opened 2015-06-30
add geocode method that use a CLRegion

pgherveou:
I am returning an array of CLPlacemark instead of just the first element here 
Its more convenient to get all the results for my use case, and it felt more natural not to modify the return type of the non promise api

Let me know if this works for you, otherwise I would just hard code this inside my app

### DIFF
--- a/Categories/CoreLocation/CLGeocoder+Promise.swift
+++ b/Categories/CoreLocation/CLGeocoder+Promise.swift
@@ -35,6 +35,15 @@ extension CLGeocoder {
             }
         }
     }
+    
+  public func geocode(addressString: String, region: CLRegion) -> Promise<[CLPlacemark]> {
+    return CLPromise { sealant in
+      geocodeAddressString(addressString, inRegion: region) { placemarks, error in
+        sealant.resolve((placemarks as? [CLPlacemark]), error)
+      }
+    }
+  }
+    
 }
 
 private var onceToken: dispatch_once_t = 0


Categories/CoreLocation/CLGeocoder+Promise.swift
```diff
@@ -35,6 +35,15 @@ extension CLGeocoder {
             }
         }
     }
+    
+  public func geocode(addressString: String, region: CLRegion) -> Promise<[CLPlacemark]> {
```
mxcl: Should be `CLRegion?` if mirroring the underlying API.

mxcl:
Thanks. Could you use the same indent-size as the rest of the file however? Thanks again.

mxcl:
Thanks. I am going to hold off on pushing this to pods though as it dictates as minor version bump.

To use this version with CocoaPods you can specify that it sources from master.

Categories/CoreLocation/CLGeocoder+Promise.swift
pgherveou: Oups ok doing it now
I am using carthage so no need to push anything i will use the commit id until next version

pgherveou:
you already merged, you still want me to change code indent and change region to inRegion?

mxcl:
I decided to merge and do it myself. It’s not very nice to expect people to do this sort of stuff IMO after they already have done some work in their own time.

pgherveou:
Sorry about this @mxcl, I was not connected when you replied 
but I was ready to do it, you were just too fast for me on this one

mxcl:
np, glad that you were happy to but it was fine for me to do it.

If the change was more severe I would have discussed it first obv.

Issue #235, opened 2015-08-09
PMKHTTPURLResponseIsJSON doesn't support application/x-javascript


jefferythomas:
I'm an sample application for a talk I'm doing. I'm using the Duck Duck Go API which returns a content type of application/x-javascript for it's JSON.

It would be nice to set the array of MIME types programmatically instead of it being hardcoded.

mxcl:
Seems to me Duck Duck Go should be passing `application/json` when they are not return JSON-P.

mxcl:
The fix will be in 2.2.2 which should be released later today.

jefferythomas:
Thanks for the quick turnaround. I also added in some code which could solve the problem in a more generic way.

PR #236, opened 2015-08-09
Added PMKSetJSONMIMETypes() to handle setting a new set of MIME types…

jefferythomas:
… for JSON responses.

Based on the issue i submitted, PMKHTTPURLResponseIsJSON doesn't support application/x-javascript #235, I created a way to set the JSON MIME types.


mxcl:
Thank you, however I don’t think this should be configurable. When will JSON mime types change in the future? It's very unlikely they will change.

For the case where for some reason the API you are using has a bad mime type, you should write your own promise. The justification for PromiseKit’s categories is they wrap existing Cocoa asynchronous patterns with promises, and I also try to add a little sugar, so for example the NSURLConnection promises also will decode JSON if they detect it.

Being able to configure what JSON *is* is beyond the scope of the library, IMO.

mxcl:
Any comments on my comment?

jefferythomas:
Hey Max,

I would agree that PromiseKit’s categories are more for simple examples then for the general messiness of production code. I supported my the Duck Duck Go behavior by adding my own `then()` clause to the promised `NSData` object. My `then()` clause turned the data into a JSON object and returned a promise for that JSON object.

It was a few lines of code which actually added another example of how to take advantage of chaining promises.

At some point in the future, when the Xcode 7/Swift 2.0 issues have been sorted out, it might be nice to enhance the `URLConnection` category with a response processing callback. I would have it do the work I did in my `then()` clause, except it would be called from every `URLConnection` promise.

With this feature, the default handing of JSON and images would simply be the default response processing callback.

Thanks for taking the time to hear me out.

Jeff

PR #245, opened 2015-08-20
Allow an array of void promises for `when`

djtarazona:
I have an array of void Promises (`Promise<Void>`) that I would like to use with `when`. It seems like `when` only supports void Promises when passed in as a variadic parameter.  Or maybe I am doing something wrong?  Anyways, this change allowed me to pass it in how I wanted.

```swift

var promises = [Promise<Void>]()

let promiseA: Promise<Void> = Promise { fulfiller, rejecter in
    //
}

let promiseB: Promise<Void> = Promise { fulfiller, rejecter in
    //
}

promises.append(promiseA)
promises.append(promiseB)

when(promises)
```

### DIFF
--- a/Sources/when.swift
+++ b/Sources/when.swift
@@ -1,6 +1,6 @@
 import Foundation.NSProgress
 
-private func when<T>(promises: [Promise<T>]) -> Promise<Void> {
+private func _when<T>(promises: [Promise<T>]) -> Promise<Void> {
     let (rootPromise, fulfill, reject) = Promise<Void>.defer()
 #if !PMKDisableProgress
     let progress = NSProgress(totalUnitCount: Int64(promises.count))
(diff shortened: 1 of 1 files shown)


djtarazona:
Actually this doesn't work. Any suggestions?

djtarazona:
Ah, got it. We just need to make `when<T>(promises: [Promise<T>]) -> Promise<Void>` public?

mxcl:
Can't merge, this will make when ambiguous with:

```
public func when<T>(promises: [Promise<T>]) -> Promise<[T]> {
    return when(promises).then(on: zalgo) { promises.map{ $0.value! } }
}
```

You should be able to pass in `[Promise<Void>]` it will produce a `Promise<[Void]>` (not `Promise<Void>` which is more useful).

A PR for:

```
public func when(promises: [Promise<Void>]) -> Promise<Void> {
    return when(promises)
}
```

would probably be ok. I'm worried it may be ambiguous however. Please run the OS X tests before submitting the PR.

I'd do the work myself but am currently legally incapable.

djtarazona:
PR updated so that `when` is not ambiguous.

Unfortunately I'm not too familiar with Carthage yet so I haven't run the tests.  I can do so later but I probably won't have time to tinker with it until another week or so.  If anyone wants to run the tests for me, that would be great!  PromiseKit compiles (and seems to work) just fine for me with this code in my iOS project.

PR #252, opened 2015-08-30
ABAddressBook category - return user denied state when it's already been denied

allen-zeng:
Currently when trying to use this category like so:

``` swift
ABAddressBookRequestAccess().then { (book: ABAddressBook) -> Void in
    // do stuff
}
.recover { error -> Void in
	let message = (error as? AddressBookError)?.localizedDescription ?? "Unknown error"

	UIAlertView(
        title: "Address book access required",
        message: message,
        delegate: nil,
        cancelButtonTitle: "OK").show()
}
```

When the application has already been denied access, The `UIAlertView` shows "Unknown error", rather than the denied message.

According to the header comments for `ABAddressBookCreateWithOptions`:
> If access to contact data is already restricted or denied, this will fail returning a NULL ABAddressBookRef with error kABOperationNotPermittedByUserError.

This change looks at the error code to ensure that it equates to `kABOperationNotPermittedByUserError`, when it does, the promise will eventually be rejected with `AddressBookError.Denied` rather than a straight up `NSError`.

### DIFF
--- a/Categories/AddressBook/ABAddressBookRequestAccess+Promise.swift
+++ b/Categories/AddressBook/ABAddressBookRequestAccess+Promise.swift
@@ -83,10 +83,16 @@ extension NSError {
     }
 }
 
-private func ABAddressBookRequestAccess() -> Promise<(Bool, ABAddressBook)> {
+private func ABAddressBookRequestAccess() -> Promise<(Bool, ABAddressBook?)> {
     var error: Unmanaged<CFError>? = nil
     guard let ubook = ABAddressBookCreateWithOptions(nil, &error) else {
-        return Promise(error: NSError(CFError: error!.takeRetainedValue()))
+        let error = error!.takeRetainedValue()
+        if CFErrorGetCode(error) == kABOperationNotPermittedByUserError {
+            let deniedResponse: (Bool, ABAddressBook?) = (false, nil)
+            return Promise(deniedResponse)
+        }
+
+        return Promise(error: NSError(CFError: error))
     }
 
     let book: ABAddressBook = ubook.takeRetainedValue()


mxcl:
Sorry, I don't understand this report.

It seems like current behavior is correct. The promise represents asking for access to the address book and is rejected if the user rejects the request.

You are making it so the promise succeeds if its premise fails.

Or do I misunderstand?

mxcl:
OK I guess I get it, since we are returning Bool we are implying the promise will be `false, nil` if the user rejects it and be an error if the task itself failed in some unexpected way.

Yes, I see. I'll have to consider what we *should* be doing. What do you think?

mxcl:
Thanks, your ticket made me look further and I think the whole file has issues. Specifically:

```swift
 public func ABAddressBookRequestAccess() -> Promise<ABAddressBook> {
     return ABAddressBookRequestAccess().then(on: zalgo) { (granted, book) -> Promise<ABAddressBook> in
         guard granted else {
             switch ABAddressBookGetAuthorizationStatus() {
             case .NotDetermined:
                 throw AddressBookError.NotDetermined
             case .Restricted:
                 throw AddressBookError.Restricted
             case .Denied:
                 throw AddressBookError.Denied
             case .Authorized:
                 fatalError("This should not happen")
             }
         }
 
         return Promise(book)
     }
 }
```

Surely this check should be in the error handler and not the then handler? We need tests for this, which may be tricky even with the new XCUITesting.

mxcl:
I think our error is assuming the combination: `false. nil` is possible.

Apple are pretty strict in this pattern for completion blocks. Thus we can assume `false, nil` is impossible.

The other error is having a custom error type. Your code is in fact incorrect above, assuming that the error will always be of type `PromiseKit.AddressBookError`, you need a more complete logic handler that checks for `NSError` too, since in your example that is what is happening.

Thus, it seems having `PromiseKit.AddressBookError` exist at all is confusing. We should either handle all errors with our own type or none. Not half and half. My bad.

I thus conclude we should handle none since we cannot know all errors Apple provide since their documentation is lacking.

mxcl:
I think this is all we should have, in fact:

```swift
import AddressBook
import CoreFoundation
import Foundation.NSError
#if !COCOAPODS
import PromiseKit
#endif

/**
 Requests access to the address book.

 To import `ABAddressBookRequestAccess`:

    use_frameworks!
    pod "PromiseKit/AddressBook"

 And then in your sources:

    import PromiseKit
*/

extension NSError {
    private convenience init(CFError error: CoreFoundation.CFError) {
        let domain = CFErrorGetDomain(error) as String
        let code = CFErrorGetCode(error)
        let info = CFErrorCopyUserInfo(error) as [NSObject: AnyObject]
        self.init(domain: domain, code: code, userInfo: info)
    }
}

public func ABAddressBookRequestAccess() -> Promise<(ABAddressBook)> {
    var error: Unmanaged<CFError>? = nil
    guard let ubook = ABAddressBookCreateWithOptions(nil, &error) else {
        return Promise(error: NSError(CFError: error!.takeRetainedValue()))
    }

    let book: ABAddressBook = ubook.takeRetainedValue()
    return Promise { fulfill, reject in
        ABAddressBookRequestAccessWithCompletion(book) { _, error in
            if error == nil {
                fulfill(book)
            } else {
                reject(NSError(CFError: error))
            }
        }
    }
}
```

Thoughts?

allen-zeng:
I'm assuming breaking changes in API doesn't really matter? This was the main reason why I didn't change more.

I don't see the need for `func ABAddressBookRequestAccess() -> Promise<ABAuthorizationStatus>` so I think removing that is probably a good idea.

You're right about needing to check for `NSError` but I also rather like the fact that I can get a reject reason within `recover`. I suppose in this particular case one could just call `ABAddressBookGetAuthorizationStatus()` within the `recover` so it doesn't really matter too much.

allen-zeng:
Either way it looks like this change isn't the way to go. Closing.

mxcl:
Reopening so I don’t forget work needs to be done.

I won’t break the API but we have a version 3.0.0 on the horizon.

cmckni3:
Should probably add support for the Contacts framework since ABAddressBook was deprecated in iOS 9.

PR #261, opened 2015-09-13
Use @rpath in DYLIB_INSTALL_NAME_BASE for OSX framework.

rupertdaniel:


mxcl:
Sounds sensible, but specifically why for your needs? May I ask?

rupertdaniel:
Ahh yeah sorry should have included a comment.

This allows for the framework to be used as an embedded framework within the app bundle. The current setting means the framework needs to be installed in /Library/Frameworks for it to be found at runtime.

I think the most common use case is the framework to be used as an embedded framework.

mxcl:
Oh that severe? Odd that the tests didn't uncover this, maybe the bug is device only?

Must be a regression versus 2.x but the xcodeproj has changed a lot.

Please merge @nathanhosselton. Ta!

PR #262, opened 2015-09-14
add @compatibility_alias for PMKAnyPromise

bgerstle:
Wasn't able to compile a project where a Swift class returned `AnyPromise` for usage in ObjC.  

I believe the generated Swift header was the problem because it had a forward declaration for `PMKAnyPromise`, which wasn't being aliased to `AnyPromise` after importing `PromiseKit.h`.  Not sure if something in `SWIFT_CLASS` has changed, but this seems to fix it.

### DIFF
--- a/Sources/Umbrella.h
+++ b/Sources/Umbrella.h
@@ -56,4 +56,6 @@ extern NSString * const PMKErrorDomain;
     @property (nonatomic, readonly) BOOL fulfilled;
     @property (nonatomic, readonly) BOOL rejected;
     @end
+
+    @compatibility_alias PMKAnyPromise AnyPromise;
 #endif


mxcl:
`AnyPromise` is `PMKAnyPromise`, per this stuff in Umbrella.h:

```objc
    SWIFT_CLASS("PMKAnyPromise")
    @interface AnyPromise : NSObject
```

Can you post the error message?

bgerstle:
Sorry I didn't do this at first, but I created an [example project](https://github.com/bgerstle/PMKInteropExample) which:

- Uses mxcl/PromiseKit swift-2.0.beta5 via Carthage
- Has a Swift `NSObject` subclass, which has a function returning `AnyPromise`
- Has an ObjC `ViewController` which tries to import the project's Swift class and call the function returning `AnyPromise`

Which results in the following error:

> /Users/bgerstle/Code/PMKInteropExample/PMKInteropExample/ViewController.m:22:50: error: property 'then' cannot be found in forward class object 'PMKAnyPromise'
    [[SwiftClassWithPromise new] doSomethingAny].then(^(id value) {
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
In file included from /Users/bgerstle/Code/PMKInteropExample/PMKInteropExample/ViewController.m:11:
/Users/bgerstle/Code/PMKInteropExample/Build/Intermediates/PMKInteropExample.build/Debug-iphonesimulator/PMKInteropExample.build/DerivedSources/PMKInteropExample-Swift.h:94:8: note: forward declaration of class here
@class PMKAnyPromise;

bgerstle:
If I [update the `Cartfile` to point at my fork](https://github.com/bgerstle/PMKInteropExample/commit/4a85c7187ba77ae552c89bd5c87b83794bb40eaa), the problem goes away.

mxcl:
Yes, this is due to a change in beta 6 in how the runtime attribute for objc classes works. Your fix is good, but I think we should just give-in here and have `AnyPromise` be `AnyPromise` in the runtime as well.

bgerstle:
So you're proposing to drop `PMKAnyPromise` entirely?  Seems like a decent trade-off of user convenience for breaking Cocoa namespace prefix conventions.

OTOH, what would happen if you just let the definition in `AnyPromise.swift` stand alone? ```

``` swift
@objc(PMKAnyPromise) public class AnyPromise
``` 

This should allow all Swift code to reference `AnyPromise` and all ObjC code to reference `PMKAnyPromise`.  I wonder if you could even add an empty `AnyPromise` swift class with an `unavailable` attribute which warns users to use `PMKAnyPromise` in ObjC instead?

nathanhosselton:
#271 fixes

PR #267, opened 2015-09-16
Use @rpath for DYLIB_INSTALL_NAME_BASE

slavikus:
For framework to be successfully embedded withing OS X app binary, the installation path for LD has to be with @rpath (just like for iOS). PMKOSX target had it set to /Library/Frameworks.


slavikus:
Yay! :) Thanks.

PR #272, opened 2015-09-17
add watchOS deployment target

### LINKED ISSUE #265: watchOS 2 support
I would love to use PromiseKit on watchOS 2. If the Pod and PromiseKit can be updated easily to support watchOS 2 as a deployment target, even better! 

[Podspec syntax for adding watchOS 2 deployment target](https://guides.cocoapods.org/syntax/podspec.html#watchos)

sahandnayebaziz:
This, along with mxcl/OMGHTTPURLRQ#8 will resolve #265

### DIFF
--- a/PromiseKit.podspec
+++ b/PromiseKit.podspec
@@ -17,6 +17,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.9'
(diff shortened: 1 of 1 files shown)


mxcl:
FFR you can amend the commit and force push the branch rather than make a cosmetic additional commit.

I don't mind, but it's good practice.

Thanks for the fix!

sahandnayebaziz:
Thank you! I didn't know you could do that. I will do that next time :)

PR #277, opened 2015-09-23
Changes for 3.0.0

nathanhosselton:


mxcl:
We should switch `report` to `error` like in the other discussion.

`report` should be kept around though for a few versions since many people are now using the switch-beta branches. Then we'll deprecate it.

PR #284, opened 2015-10-02
Swift 2.0 beta5

kdubb:
• Update MessageUI naming
• Make PMKAlertController error and cancellation conform to new standards


kdubb:
I incorrectly pushed to this branch which compounded the original changes I was trying to get integrated.  I will break them out into their own branches/PRs for clarity.

Issue #241, opened 2015-08-17
CLLocation promise should ignore initial kCLErrorLocationUnknown error


tgaul:
As documented by Apple here: https://developer.apple.com/library/ios/documentation/CoreLocation/Reference/CLLocationManagerDelegate_Protocol/index.html#//apple_ref/occ/intfm/CLLocationManagerDelegate/locationManager:didFailWithError:

> The location manager calls this method when it encounters an error trying to get the location or heading data. If the location service is unable to retrieve a location right away, it reports a kCLErrorLocationUnknown error and keeps trying. In such a situation, you can simply ignore the error and wait for a new event. If a heading could not be determined because of strong interference from nearby magnetic fields, this method returns kCLErrorHeadingFailure.

In the implementation of CLLocationManager's `promise` API (both ObjC and Swift, from what I could tell — I'm still on PromiseKit 1.6, but this also seems to still be this way in 2.0), it resolves the promise on the first call to the `locationManager:didFailWithError:` API, which means it hasn't waited long enough for a successful location. And because this causes the object to stop updating location, it's very possible that a client trying to handle this itself in a `catch` block by returning another location promise would just trigger the same situation again.

A better way to handle this would seem to be to check for a `kCLErrorDomain` / `kCLErrorLocationUnknown` error, and in that case, simply not resolve the promise and wait for a "real" result or error.

mxcl:
Thanks, and PR welcome, I cannot fix this. Not due to lack of desire, I assure you. Situation makes it impossible.

tgaul:
Will look into it, when I get out from under deadline. You want a PR only for 2.0, or also 1.6? (Also, congrats on the new "situation", if it's what it looks to be.)

mxcl:
2.0 only is fine by me. Thanks and thanks :)

PR #286, opened 2015-10-07
fix https://github.com/mxcl/PromiseKit/issues/241

pgherveou:
this should fix the issue described here 
https://github.com/mxcl/PromiseKit/issues/241


pgherveou:
oups bad commit, sorry about that resubmitting

PR #287, opened 2015-10-07
fix https://github.com/mxcl/PromiseKit/issues/241

pgherveou:


pgherveou:
fixing https://github.com/mxcl/PromiseKit/issues/241 ...

pgherveou:
I think I was having issue running this code in the simulator, but it might be cleaner to remove the delay and wait for a new event

nathanhosselton:
I worry that the delay may preemptively reject the promise, cancelling out the advantage of ignoring the error in the first place. Unless you have tested that this does not happen.

PR #304, opened 2015-10-26
Allow firstly to take a queue

kdubb:
Allows nice syntax like ```firstly(on: mainQueue) {```


mxcl:
This is what `dispatch_promise` is for.

PR #303, opened 2015-10-26
Alert controller cancellation

kdubb:
PMKAlertController "Cancel" style actions now cancel the promise (currently it's reversed).  It also makes the Error enum conform to CancellableErrorType as required.

PR #310, opened 2015-10-31
Fixes PMKAlertController addActionWithTitle fulfill/reject bug fix

MrAdamBoyd:
Before, promises were only being fulfilled when the "Cancel" button was hit. All other button hits were triggering `self.reject()`. This is now reversed. All actions added now fulfill promises and the Cancel button triggers `self.reject()`. 1 character change.


MrAdamBoyd:
Actually, realized there is already a pull request for this [here](https://github.com/mxcl/PromiseKit/pull/303).

PR #319, opened 2015-11-12
Create typed .always() methods

andreterron:
Improve .always by creating two new typed methods:
```swift
.always(on: dispatch_queue_t, body: () throws -> U) -> Promise<U>
.always(on: dispatch_queue_t, body: () throws -> Promise<U>) -> Promise<U>
```

Similar to:
```swift
.then(on: dispatch_queue_t, body: (T) throws -> U) -> Promise<U>
.then(on: dispatch_queue_t, body: (T) throws -> Promise<U>) -> Promise<U>
```

But it will be executed whether it failed or succeeded

### DIFF
--- a/Sources/Promise.swift
+++ b/Sources/Promise.swift
@@ -437,6 +437,54 @@ public class Promise<T> {
         }
     }
 
+    /**
+     The provided closure is executed when this Promise is resolved.
+
+         somePromise().then {
+            //…
+         }.always { () -> Int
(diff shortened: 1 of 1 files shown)


mxcl:
`always` cannot change the promise chain’s value this is deliberate. What if the chain errors?

```swift
foo.then {
    throw Error.foo`
always {
    return 1  // cancels error
}
```

PR #325, opened 2015-11-18
Updates for TVos.

tjdhome:
I've added a new target for TVos. Seems to run well in a very demanding (up to dozens of promise nodes in a complicated tree). Thanks so much for this package!!!

### DIFF
--- a/PromiseKit.xcodeproj/project.pbxproj
+++ b/PromiseKit.xcodeproj/project.pbxproj
@@ -7,6 +7,30 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		3EABFED61BF8D959006F1F50 /* PromiseKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3EABFECC1BF8D959006F1F50 /* PromiseKit.framework */; };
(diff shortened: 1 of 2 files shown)


sebastien-de-saint-florent:
Hi,
I would need also PromiseKit to be tvos compatible.
@tjdhome, is it normal you haven't updated the podspec file to include something like s.tvos.deployment_target = '9.0' ?

tjdhome:
Normal for me :-) I do not use cocoa pods. Too many unintended consequences of updates!

 - tjd



On Nov 27, 2015, at 2:37 PM, "S?bastien De Saint Florent" <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi,
I would need also PromiseKit to be tvos compatible.
@tjdhome<https://github.com/tjdhome>, is it normal you haven't updated the podspec file to include something like s.tvos.deployment_target = '9.0' ?

-

nathanhosselton:
PromiseKit should now have proper tvOS support.

@tjdhome I want apologize that your contribution was not utilized in said support, but would still like to thank you for your time and effort towards making PromiseKit better. It is always appreciated.

PR #329, opened 2015-11-19
Fix backwards PMKAlertController Canceled logic

nevans:
It was rejecting the non-cancels and fulfilling the cancels!


swaffoja:
+1

PR #343, opened 2015-12-11
Fix #342. Use proper instance method…

### LINKED ISSUE #342: Nonexistent method (promiseWithResolver:) referred to in documentation.
The class method `promiseWithResolver:` has three references in documentation, but the correct use is the instance method `initWithResolver:`.

thaterikperson:
so people can find their way a little easier when reading through the documentation on 
```objc
- (AnyPromise *)promiseViewController:(UIViewController *)vc animated:(BOOL)animated completion:(void (^)(void))block;
```

### DIFF
--- a/Categories/UIKit/UIViewController+AnyPromise.h
+++ b/Categories/UIKit/UIViewController+AnyPromise.h
@@ -51,7 +51,7 @@
     }
  
     - (void)viewDidLoad {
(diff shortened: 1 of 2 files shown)


thaterikperson:
Addressed feedback from @nathanhosselton in #342.

PR #346, opened 2015-12-30
join promises with Swift's array argument

paweljankowski:
This PR implements join method for Swift's array argument. The implementation is the same as implementation for the variadic function.

### DIFF
--- a/Sources/join.swift
+++ b/Sources/join.swift
@@ -42,3 +42,30 @@ public func join<T>(promises: Promise<T>...) -> Promise<[T]> {
         }
     }
 }
+
+public func join<T>(promises: [Promise<T>]) -> Promise<[T]> {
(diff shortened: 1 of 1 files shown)


nathanhosselton:
Thanks for the contribution! I think this is a good idea but the code duplication isn't ideal. If the implementations can be identical, as it seems they can, I would recommend simply having one method call the other. Though I would imagine you thought of this so perhaps there's some reason why this isn't trivially possible.

randomiser:
+1 

Until this gets resolved:

https://bugs.swift.org/browse/SR-128

We need to be able to dynamically provide a list of promises at runtime.

nathanhosselton:
Resolved by #374. But nevertheless, your efforts are very much appreciated @paweljankowski.

PR #362, opened 2016-02-15
Added CoreBluetooth+Promise category

aronse:
Added a `promise()` method on `CBCentralManager` to retrieve an initialized CBCentralManager object easily. 
This can be useful in case one needs to check the device's bluetooth state.

Example:
```swift
CBCentralManager.promise().then { manager -> Void in
  if manager.state == .PoweredOff {
    // ask user to turn his bluetooth on
  }
}
```

### DIFF
--- a/Categories/CBCentralManager+Promise.swift
+++ b/Categories/CBCentralManager+Promise.swift
@@ -1,0 +1,44 @@
+import Foundation
+import CoreBluetooth
+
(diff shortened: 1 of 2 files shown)


mxcl:
Merged: d1f77c76c2852dc24a79b92c1596ec8c9dc5a1ac

PR #366, opened 2016-02-25
Give a typealias for pendingPromise result.

dtartaglia:
I had to declare a variable to hold the result of a pendingPromise call and the type declaration is very long. This will make it a little easier to type... Not sure about the name `ExpandedPromise` though. Maybe it should be `PendingPromise`?

### DIFF
--- a/Sources/Promise.swift
+++ b/Sources/Promise.swift
@@ -193,9 +193,9 @@ public class Promise<T> {
        2) A function that fulfills that promise
        3) A function that rejects that promise
     */
-    public typealias ExpandedPromise = (promise: Promise, fulfill: (T) -> Void, reject: (ErrorType) -> Void)
+    public typealias PendingPromise = (promise: Promise, fulfill: (T) -> Void, reject: (ErrorType) -> Void)
     
-    public class func pendingPromise() -> ExpandedPromise {
+    public class func pendingPromise() -> PendingPromise {
         var fulfill: ((T) -> Void)!
         var reject: ((ErrorType) -> Void)!
         let promise = Promise { fulfill = $0; reject = $1 }
--- a/Sources/Promise.swift
+++ b/Sources/Promise.swift
@@ -193,7 +193,9 @@ public class Promise<T> {
        2) A function that fulfills that promise
        3) A function that rejects that promise
     */
-    public class func pendingPromise() -> (promise: Promise, fulfill: (T) -> Void, reject: (ErrorType) -> Void) {
+    public typealias ExpandedPromise = (promise: Promise, fulfill: (T) -> Void, reject: (ErrorType) -> Void)
+    
+    public class func pendingPromise() -> ExpandedPromise {
         var fulfill: ((T) -> Void)!
         var reject: ((ErrorType) -> Void)!
         let promise = Promise { fulfill = $0; reject = $1 }


nathanhosselton:
I believe the intention with `pendingPromise()` was for the user to hold references to each of the values in the returned tuple individually rather than the tuple itself. May I ask what your needs are that you are retaining the tuple? I agree that a typealias would make sense here if that's the case.

dtartaglia:
I am holding the result of a pendingPromise call as a property of a class, but I can't create the actual promise at initialization time. As a consequence I have the following line of code in my class:

`private var pendingAdd: (promise: Promise<(name: String, date: NSDate)>, fulfill: ((name: String, date: NSDate)) -> Void, reject: (ErrorType) -> Void)?`

which is pretty ugly. With the addition of the typealias I could write this instead:

`private var pendingAdd: Promise<(name: String, date: NSDate)>.ExpandedPromise?`

Otherwise, I am using it exactly as intended.

nathanhosselton:
This makes sense. But I'm assuming when you need to call `fulfill` and `reject` or access the promise you then do `pendingAdd.fulfill()` etc. My question was why not create separate properties for each of the values in the tuple (the `promise`, `fulfill`, and `reject`) as I believe this was the intention and why a typealias wasn't provided.

Hopefully I'm not coming off as trying to be difficult because that's not my intention. Just trying to understand your needs before moving forward.

dtartaglia:
Hmm... I'm not sure how I could do that... For example:

    class Bar {
	    var (promise, fulfill, reject): (promise: Promise, fulfill: (Int) -> Void, reject: (ErrorType) -> Void)
	    // Create a promise in some method.
    }

I can't do the above. The tuple needs to be initialized in the init method (which, as I said, I can't do.)

    class Bar {
	    var (promise, fulfill, reject): (promise: Promise, fulfill: (Int) -> Void, reject: (ErrorType) -> Void)?
	    // Create a promise in some method.
    }

The above doesn't work either. I get the error "tuple pattern cannot match values of the non-tuple type."

    class Bar {
	    var (promise, fulfill, reject): (promise: Promise?, fulfill: ((Int) -> Void)?, reject: ((ErrorType) -> Void)?)
    }

The above allows me to define the variable, but then I won't be able to assign the return of pendingPromise() to it.

And in any case, all of these constructs are very messy.

Maybe I've missed something fundamental and I should be doing what I am trying to do in a completely different way?

nathanhosselton:
What I meant was:

```swift
class Foo {
    var promise: Promise<(String, NSDate)>?
    var fulfill: ((String, NSDate) -> Void)?
    var reject: ((ErrorType) -> Void)?

    func startPromise() {
        (promise, fulfill, reject) = Promise<(String, NSDate)>.pendingPromise()
    }
}
```
Which I think is a little cleaner if only because it breaks the declaration down into smaller pieces that are easier to digest. However, you're onto something in that there appears to be some sort of Swift shortcoming that prevents the mapping of tuples when the references being assigned to are declared Optional. So in order for this to work, I had to do:

```swift
    func startPromise() {
        let t = Promise<(String, NSDate)>.pendingPromise()
        promise = t.promise
        fulfill = t.fulfill
        reject = t.reject
    }
```

Which is slightly better than having to access the values via the tuple whenever you need them, though I'll admit not by much. However, I can't see how the introduction of the typealias will skirt this issue since you'll still be declaring the property as optional and, more than that, the values within won't be typed. Have you tested? I have attempted to but am having some Xcode problems at the moment.

dtartaglia:
That's an interesting construct. I hadn't thought of it. As for my testing, I am currently using the following in my code and it works fine:

    extension Promise {
        typealias PendingPromise = (promise: Promise<T>, fulfill: (T) -> Void, reject: (ErrorType) -> Void)
    }

    class Foo {
    	private var pendingAdd: Promise<(name: String, date: NSDate)>.PendingPromise?
    }

I'm happy to leave it in my code but it seemed to me like something that should be in the library.

nathanhosselton:
Ah I see. I misunderstood how typealiases worked in Swift. The typing is not an issue with the typealias, but the optional mapping of the tuples still is. I receive the same error with the typealias property as I received before and that I assume you received with your third example in your message before last. I think this would be a valuable addition to PK if we can make it work, but I don't understand why we're having different experiences at the moment.

nathanhosselton:
I apologize, I realize my last response wasn't entirely constructive. I implemented the same as in your prior example and all is fine until I actually assign to the property the return value of `pendingPromise()`:
```swift
self.pendingAdd = Promise<(String, NSDate)>.pendingPromise()
```
On that line, I receive the tuple mapping error I've mentioned previously that I believe to be a Swift issue:
```
Cannot assign value of type '(promise: Promise<(String, NSDate)>, fulfill: ((String, NSDate)) -> Void, reject: (ErrorType) -> Void)' (aka '(promise: Promise<(String, NSDate)>, fulfill: (String, NSDate) -> (), reject: ErrorType -> ())') to type '(promise: Promise<(name: String, date: NSDate)>, fulfill: (name: String, date: NSDate) -> Void, reject: (ErrorType) -> Void)?'
```
Just to be safe I also updated the PK source to match your PR but had the same result. Again, I'm not sure why our results would conflict with the same apparent code. Let me know if you have any ideas.

dtartaglia:
No problem Nathan. Two things I noticed that are, or at least may be. different between what I have and what you have...

I am doing this in an @objc class (specifically a UIViewController subclass.)

`private var pendingAdd: Promise<(name: String, date: NSDate)>.PendingPromise?`

and I am assigning to it like this:

	func waitForAdd() -> Promise<(name: String, date: NSDate)>  {
		pendingAdd = Promise<(name: String, date: NSDate)>.pendingPromise()
		return pendingAdd!.promise
	}

So it could be that it only works in an @objc class, or it could have something to do with the fact that I used named parameters in my tuple. ?

I just tried removing the named parameters, and my code still worked... Hmm...

nathanhosselton:
OK, I think I have this all figured out, the result being that your `typealias` inclusion definitely has value.

The reason it wasn't working for me was an error on my part. I had included the parameter names in the property declaration but weren't including them when I was actually assigning to the property. It works either way, as long as there's consistency with the parameter names. And the `typealias` for the `pendingPromise()` return value does help a lot by cutting out all of the complex syntax when assigning directly to an optional tuple property as you are doing:
```swift
class ViewController: UIViewController {
    private var pendingAdd: Promise<(String, NSDate)>.PendingPromise?

    func waitForAdd() -> Promise<(String, NSDate)>  {
        pendingAdd = Promise<(String, NSDate)>.pendingPromise()
        return pendingAdd!.promise
    }
}
```
In contrast, this is the best solution I could find for holding separate properties for each of the returned tuple's values:
```swift
class ViewController: UIViewController {
    var promise: Promise<(String, NSDate)>?
    var fulfill: (((String, NSDate)) -> Void)?
    var reject: ((ErrorType) -> Void)?

    func waitForAdd() -> Promise<(String, NSDate)>  {
        let (p, f, r) = Promise<(String, NSDate)>.pendingPromise()
        promise = p
        fulfill = f
        reject = r
        return promise
    }
}
```
Which I don't think is bad by any means, but it's certainly not as clean if the preference is for a single property.

The reason `(promise, fulfill, reject) = Promise<(String, NSDate)>.pendingPromise()` doesn't work is because the properties are optionals and the values in the returning tuple are not, therefore they do not map and must be explicitly assigned outside of the tuple. So when it is not possible to call `.pendingPromise()` during the calling class' initialization one of the above patterns must be used. And I think it makes sense to provide the `typealias` for those who would prefer the single-property pattern.

So that brings us to naming. I agree that "PendingPromise" is preferable to "ExpandedPromise" since it is more specific to its related method, and even though it may be a little confusing to name the `typealias` for the return value of a method the same as the method itself, it kind of makes sense. So if you want to rename it on your fork I will merge. Or I'm happy to pull the PR locally and make the change. I'm going to add some documentation anyway. Up to you.

dtartaglia:
Also, if the user needs to hold more than one pending promise, then forcing them to keep three variables per promise and keep straight which promise, fulfill, rejects go together would be a huge pain.

nathanhosselton:
Good point. Hadn't even thought of that yet.

Issue #326, opened 2015-11-18
Need non-variadic Swift race and join methods


nevans:
As far as I can tell, Swift's type system won't let me apply arrays onto variadic parameters.

````swift
public func race<T>(promises: Promise<T>...) -> Promise<T> {
    return Promise(sealant: { resolve in
        for promise in promises {
            promise.pipe(resolve)
        }
    })
}
````
could perhaps be rewritten as
````swift
public func race<T>(promises: Promise<T>...) -> Promise<T> {
    return race(promises)
}

public func race<T>(promises: [Promise<T>]) -> Promise<T> {
    return Promise(sealant: { resolve in
        for promise in promises {
            promise.pipe(resolve)
        }
    })
}

````

nevans:
also for `join`

tmandry:
+1

nathanhosselton:
Seems fine. PR welcome.

randomiser:
Can we also have the join changed so that it can be chained along with the other functions like "then".

ie:

firstly {

}.then {

}.join {

}.always {

}.error {

}

nathanhosselton:
Resolved by #374

nathanhosselton:
@randomiser I think this topic probably deserves it's own discussion thread. If you would like to make a proposal of sorts we can definitely consider such a feature.

PR #374, opened 2016-03-10
  add array-based overloads for variadic parameterized functions (join, race)

feighter09:
solves issue https://github.com/mxcl/PromiseKit/issues/326

### DIFF
--- a/Sources/join.swift
+++ b/Sources/join.swift
@@ -17,6 +17,12 @@ import Dispatch
  - Returns: A new promise that resolves once all the provided promises resolve.
 */
 public func join<T>(promises: Promise<T>...) -> Promise<[T]> {
+    return join(promises)
+}
+
+public func join<T>(promises: [Promise<T>]) -> Promise<[T]> {
+    guard !promises.isEmpty else { return Promise<[T]>([]) }
+  
     var countdown = promises.count
     let barrier = dispatch_queue_create("org.promisekit.barrier.join", DISPATCH_QUEUE_CONCURRENT)
     var rejected = false
--- a/Sources/race.swift
+++ b/Sources/race.swift
@@ -1,3 +1,4 @@
+import Foundation.NSError
 
 /**
  Resolves with the first resolving promise from a set of promises.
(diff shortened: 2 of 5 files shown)


mxcl:
The reason this wasn't there is: it crashed certain versions of the Swift compiler.

Probably we're safe now, but well… just a warning.

nathanhosselton:
Will test before merging.

feighter09:
Ah the pains of swift we take for the good, huh? 
That's understandable, and I don't have every version of Xcode ever to test em all, so do with this what you wish

feighter09:
Although `wait` has had this since August of last year, so I think it'll be alright
https://github.com/mxcl/PromiseKit/commit/a2718b2d3ef2aedab42bca376aec45ee343ed9a4#diff-65c2204ae94d363ee89e320cf7f237d3

feighter09:
Don't merge, think I found an issue. Will also include tests from #375 in final

nathanhosselton:
Not sure if it's proper to take someone else's code. Push up your fix though. I'll see if @rbobbins would like to resubmit her PR with only the tests. Or we'll figure out some other arrangement. You both have contributed so I think both of you should be credited for your respective work.

feighter09:
@rbobbins do you want to replay your commit on top of mine? I'll make you a contributor to my fork if you'd like

Sources/join.swift
```diff
@@ -17,6 +17,12 @@ import Dispatch
  - Returns: A new promise that resolves once all the provided promises resolve.
 */
 public func join<T>(promises: Promise<T>...) -> Promise<[T]> {
+    return join(promises)
+}
+
+public func join<T>(promises: [Promise<T>]) -> Promise<[T]> {
+    guard !promises.isEmpty else { return Promise<[T]>([]) }
```
feighter09: I'm uncertain what the contract here is - should an empty array of promises to `join` immediately resolve or should an "invalid use" error be produced like below?

Sources/race.swift
```diff
@@ -12,6 +13,16 @@
...
 public func race<T>(promises: Promise<T>...) -> Promise<T> {
+    return try! race(promises)  // race only throws when the array param is empty, which is not
+                                // possible from this variadic paramater version
+}
+
+public func race<T>(promises: [Promise<T>]) throws -> Promise<T> {
+    guard !promises.isEmpty else {
+        let message = "Cannot race with an empty list of runners (Promises)"
+        throw NSError(domain: PMKErrorDomain, code: PMKInvalidUsageError, userInfo: ["messaage": message])
+    }
```
feighter09: Not sure how this should be handled, I'm unable to make this immediately resolve in the case of an empty array because we cannot instantiate a `T` for a use like `return Promise<T>(value: T())`. This forces me to `throw` an error or return an optional promise `Promise<T>?`, but this seemed less desirable.

mxcl: IMO it's an error to race an empty array. `when` can logically work with an empty array since whatever the count it completes when they all complete, if there are no promises to complete, it is already completed. With `race` *one* must succeed so if you feed none, it will fail.

feighter09:
added join tests and array-based race tests. Anything else preventing this from merge?

nathanhosselton:
Will review at length ASAP. Though at first glance everything seems good.

nathanhosselton:
Thanks a bunch to @feighter09  and @rbobbins both.

PR #375, opened 2016-03-10
Updated race to support array of promises, not just varargs

rbobbins:


Sources/race.swift
```diff
@@ -11,10 +11,14 @@
  - Returns: A new promise that resolves when the first promise in the provided promises resolves.
  - Warning: If any of the provided promises reject, the returned promise is rejected.
 */
-public func race<T>(promises: Promise<T>...) -> Promise<T> {
+public func race<T>(promises: [Promise<T>]) -> Promise<T> {
     return Promise(sealant: { resolve in
```
feighter09: There's an issue here if the array param is empty, the promise will never fulfill. I've handled this in #374 but could really use your tests, would you like to merge?

nathanhosselton:
Duplicate of #374 

I apologize that your effort went unthanked this time around but please continue to contribute!

Issue #311, opened 2015-11-01
PromiseKit 1.6.x - Support for tvOS and WatchOS


mofirouz:
I'm developing a library that is dependant on AFN and PromiseKit. The latest AFN (3.0.0-beta) has support for watchOS and tvOS, and so does PromiseKit 2.x/3.x. 

Is there anyway that we can add support for the said targets in PMK 1.6 please?

Thanks!

staminajim:
I can confirm that by adding 
`s.watchos.deployment_target = '2.0'`
to the pod spec with the v1.6.0 tag that using PromiseKit 1.6 works on watchOS.

I could make a pull request, but I'm not sure how you guys plan to keep maintaining 1.6.x as it looks like it's currently all a linear history in master?

mofirouz:
Thanks @staminajim - It seems like WatchOS is working, but tvOS is still no go.

staminajim:
@mofirouz does it work if you add 
`s.tvos.deployment_target = '9.0'` to the 1.6.0 pod spec?

I don't have any tvOS targets set up to quickly try that.

mofirouz:
Sadly I don't have a target that will enable that as well. I've added support for it in the [library](http://github.com/heroiclabs/heroiclabs-sdk-ios) that I'm building for whenever PromiseKit adds that in.

On 24 February 2016 at 10:05:12, James V (notifications@github.com) wrote:

@mofirouz does it work if you add
s.tvos.deployment_target = '9.0' to the 1.6.0 pod spec?

I don't have any tvOS targets set up to quickly try that.

jasonzurita:
I had a need to use Heroic Labs with tvOS, so I figured I would provide an update:

I locally copied both Heroic Labs and PromiseKit (v1.6), modified the pod spec files to include the tvOS target, and the pods successfully installed!

mxcl:
PR welcome, testing this is the tricky part, can advise how.

mofirouz:
@mxcl I would but I honestly don't know where/which branch the code for the 1.6.x release is - similar to what @staminajim said above. 

If you could advise on the git strategy for this project, I'll open a PR asap.

mxcl:
There is no branch, check out the [tag](https://github.com/mxcl/PromiseKit/releases/tag/1.6.0).

    git checkout -b 1.6.0

staminajim:
[...]

PR #380, opened 2016-03-15
tvOS and watchOS Support for PMK 1.x 

mofirouz:
This PR references issue #311. 

- Add necessary pod configs
- Updated XCode project to 1.7.0

Github didn't let me create a PR which had the 1.6.0 tag as it's base so I had to select master. To compare, please follow this https://github.com/heroiclabs/PromiseKit/commit/315e2f84b0d14721dcd0768ebbc764534bf00ff3


mxcl:
This merges master. So that's not what we want.

I guess it didn't occur to me that you can't PR onto a tag. We'll need to create a branch. We never intended not to, just when we split 1.x off we didn't need one at the time.

mofirouz:
@mxcl As you said, you can't create a PR based off a tag, but you can still commit the code without a branch to a new tag - the catch is only the owners can do that (because you are pushing directly to a commit/tag not to a PR).

I personally vote that for the sake of clarity for potentially future PRs, we create a branch for 1.x branch?

nathanhosselton:
Agreed on the new branch. Not sure what to call it though. `promisekit-1.x` would work fine, but it's purpose would not be self evident. Something like `promisekit-objc`? Open to suggestions before creation.

nathanhosselton:
I think I like `1.x` more despite my own criticisms. Will go with that unless there's objections.

mofirouz:
Thanks a lot @nathanhosselton - Let me know when I should rebase the code for easier merge for you guys.

nathanhosselton:
The new branch is now available at [`legacy-1.x`](https://github.com/mxcl/PromiseKit/tree/legacy-1.x). Sorry for the delay.

mofirouz:
I'm closing this in favor of a new PR that is based on the legacy-1.x branch.

Issue #377, opened 2016-03-13
Compiler Error with Standalone Distribution v2.2.2


DominikDeak:
*Edit: This issue also applies to v3.0.3 (non-standalone version).*

I get the following build error when I import `<PromiseKit/PromiseKit.h>`:

> /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/include/dispatch/queue.h:25:2: "Please #include \<dispatch/dispatch.h\> instead of this file directly."

The error originates from Line 1 in the `PromiseKit.h` header, where libdispatch related headers is imported: `#import <dispatch/queue.h>`

Replacing that with `#import <dispatch/dispatch.h>` fixes the problem.

mxcl:
Platform, Xcode version?

DominikDeak:
Mac Pro (Mid 2010)
El Capitan 10.11.3 (15D21)
Xcode Version 7.2.1 (7C1002)

mxcl:
This is pretty much the same platform/Xcode as everyone else, so I wonder why nobody else is getting this error?

tonyarnold:
This appears to be due to our use of Objective-C++ (in this instance — not sure if you want to share the sample project here @DominikDeak?). 

@mxcl are you open to a PR for this issue, given that the fix shouldn't have any adverse effects on plain old Objective-C?

DominikDeak:
> in this instance — not sure if you want to share the sample project here @DominikDeak? 

Yes, I was going to, but held off posting it here until you  confirmed the build error independently. 


Minimal test project is available here:
https://www.dropbox.com/s/oyqguv9im3mksgq/Foo.zip?dl=0

1. Unzip and open `Foo.xcworkspace`.
2. Optional: Run `carthage update` in the terminal under the `Foo` directory and rebuild PromiseKit. However, for convenience, I have included the PromiseKit binaries in the Zip, so this step may be unnecessary.
3. In Xcode open `Test.mm` to see where PromiseKit  header is imported.
4. Build, and you should see the error *"Please #include \<dispatch/dispatch.h\> instead of this file directly."*
5. In `Test.mm` importing ` <dispatch/dispatch.h>` before PromiseKit header fixes the problem.
6. Renaming  `Test.mm` to `Test.m` also fixes the problem. 

> so I wonder why nobody else is getting this error?

I'm speculating two possible causes:

[...]

PR #383, opened 2016-03-17
Header fix for issue #377.

DominikDeak:
Pull request that addresses issue  #377. 

In response to https://github.com/mxcl/PromiseKit/issues/377#issuecomment-196921790 I've implemented the fix that will conditionally import `<dispatch/dispatch.h>` for Objective-C++ projects, while still only import `<dispatch/queue.h>` for vanilla Objective-C projects.


nathanhosselton:
Thanks for your and @tonyarnold's diligence in researching and correcting this.

DominikDeak:
No problem.

PR #386, opened 2016-03-22
Update to Swift 2.2

leoMehlig:
Update all code to work with Xcode 7.3

I've also update the test. This required `carthage update` to be run with the `--no-use-binaries` argument, in order for carthage to not use the prebuild binary from OHHTTPStubs.

I couldn't get the `test_fulfills_with_image` and the `test_fulfills_with_data` succeed, but I assume that is because of my setup.

### DIFF
--- a/Tests/CorePromise/ErrorUnhandler.test.swift
+++ b/Tests/CorePromise/ErrorUnhandler.test.swift
@@ -153,7 +153,7 @@ class ErrorHandlingTests_Swift: XCTestCase {
             case Test
         }
 
-        Promise<Void> { _, reject in
+        _ = Promise<Void> { _, reject in
             after(0.1).then {
                 throw Error.Test
             }.error { err in
--- a/Tests/Promises:A+ Specs/0.0.0.swift
+++ b/Tests/Promises:A+ Specs/0.0.0.swift
@@ -64,20 +64,22 @@ extension XCTestCase {
 
 /////////////////////////////////////////////////////////////////////////
 
-    private func mkspecify<T>(numberOfExpectations: Int, generator: () -> T, body: (Promise<Int>, [XCTestExpectation], T) -> Void)(_ desc: String, feed: (T) -> (Promise<Int>, () -> Void)) {
-        let floater = expectationWithDescription("")
-        later(2, floater.fulfill)
-
-        let value = generator()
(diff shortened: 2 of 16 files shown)


Tests/CorePromise/ErrorUnhandler.test.swift
```diff
@@ -153,7 +153,7 @@ class ErrorHandlingTests_Swift: XCTestCase {
             case Test
         }
 
-        Promise<Void> { _, reject in
+        let _ = Promise<Void> { _, reject in
```
mxcl: `_ =` is sufficient.

Tests/Promises:A+ Specs/0.0.0.swift
```diff
@@ -12,11 +12,11 @@ func later(ticks: Int = 1, _ body: () -> Void) {
 
 
 extension XCTestCase {
-
+    
     func testFulfilled(numberOfExpectations: Int  = 1, body: (Promise<Int>, [XCTestExpectation], Int) -> Void) {
-
+
```
mxcl: whitespace changes… let’s not.

leoMehlig:
Is it ok that the PR contains multiple commits and does the last commit correct all the whitespace?

mxcl:
If you look at the diff you can see there are many cases where no code has changed, just whitespace. Some of the changes appear to be spaces -> tabs.

mxcl:
Appears good! 👍🏻

mxcl:
Ideally, you would squash these commits. You can force push back to your branch to fix the PR.

leoMehlig:
Done!

mxcl:
Many thanks! 🎆@nathanhosselton please merge!

mxcl:
Though check for API breakage, we may need to bump the major version again…

leoMehlig:
Pretty sure it doesn't break any API. All changes (++, --) are in functions bodies and the currying only was used with a private function.

nathanhosselton:
Sorry for not merging this yet. Been caught out of town in a snowstorm. In my initial review I didn't notice any API breaking changes but I would like to give it a second look to be certain before merging. Probably later today.

nathanhosselton:
Will bump the minor version and get this into CocoaPods by tomorrow evening.

Issue #388, opened 2016-03-24
Downcasting AnyPromise to Promise<T>


kominak:
In my recent effort to convert a PromiseKit 1 app (hybrid ObjC+Swift) to use PromiseKit 3, I had to deal with type-less `AnyPromises` coming from ObjC part of the app in the newer Swift part. In order to use the full advantage of typed `Promise<T>`, I came up with this downcasting approach:

```objc
// Legacy ObjC code
@interface Backend
+ (AnyPromise *)fetchUser;
@end
```

Downcasting in Swift:

```swift
Backend.fetchUsers()
.toPromise(User) // downcast to Promise<User>
.then { (user: User) -> Void in
    // ...
}
```

Of course, I can downcast to any type:
```swift
.toPromise(User?)
// or
.toPromise([User])
// etc.
```

All I had to do was adding this category:

```swift
extension AnyPromise {
    func toPromise<T>(type: T.Type) -> Promise<T> {
        return self.then { (value: AnyObject?) -> T in
            if let value = value as? T {
                return value
            }
            throw PromiseError.CastingError
        }
    }
}
```

Do you consider this a generic enough approach? Or is there any better approach, perhaps already supported by PromiseKit 3? I couldn't find anything like that, but I'd be glad if this works out of box. Of course I can prepare a PR, etc., but I'd like to hear your opinions.

Thanks for considering this addition.

nathanhosselton:
> In order to use the full advantage of typed Promise<T>, I came up with this downcasting approach

Can you provide some examples where you're using the converted promise in a way that you could not with its AnyPromise counterpart? This may help me figure out what the best solution here is. I've personally never needed to cast an AnyPromise to Promise<T>, and don't recall it ever coming up before, but that doesn't mean there isn't a valid use case for it.

kominak:
I encountered several annoying obstacles when working with `AnyPromise` in Swift, so this approach brought in several advantages (not only syntactic ones), which unifies the access to both types of promises in Swift code.

* AFAIK there's no way to call `catch`, `recover` nor `finally` on `AnyPromise` in Swift. I had to add a dummy call like `.then { _ -> Void in }`

* I found it nicer to use `.toPromise(Type)` than the code below (although I acknowledge that the second example is perhaps more clear and obvious - but requires a lot more writing):
```swift
.then { value in
    guard let value = value as? Type else { throw SomeError }
}
```
[...]

PR #389, opened 2016-03-24
Downcasting AnyPromise to Promise<T> using .toPromise(T)

kominak:
The motivation behind this addition is described here: https://github.com/mxcl/PromiseKit/issues/388

When working with legacy ObjC parts returning `AnyPromises`, it's nice to have an option to convert to a typed `Promise<T>`. This PR enables just this:

```swift
LegacyBackend.fetchUser()   // AnyPromise
.toPromise(User)            // Promise<User>
.then { user in
    // Now we can work with user: User
}
```

I found this very useful when working with large chunks of ObjC code that is otherwise fine and isn't worthy of rewriting to Swift just to get the typed promises.

When the downcasting fails, an error is thrown.

### DIFF
--- a/Sources/AnyPromise.swift
+++ b/Sources/AnyPromise.swift
@@ -213,6 +213,20 @@ import Foundation.NSError
         })
     }
 
+    /**
+     - Returns: downcasted (typed) `Promise<T>`.
+     Throws `CastingError.CastingAnyPromiseFailed(T)` if self's value cannot be downcasted to the given type.
+     Usage: `anyPromise.toPromise(T).then { (t: T) -> U in ... }`
+    */
+    public func toPromise<T>(type: T.Type) -> Promise<T> {
+        return self.then { (value: AnyObject?) -> T in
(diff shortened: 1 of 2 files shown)


nathanhosselton:
Sorry for the delay on getting to this. I've been reviewing it off and on since submission, and have been trying different approaches to make this better, but I can't find any I like. This seems to be the best implementation we can provide at this time (at least that I have found).

I would however like to make some adjustments to the documentation and maybe add some tests. I'll simply amend/squash them onto your commit before merging if you approve.

mxcl:
Cool. I think the only reason this was not there previously was that it crashed Swift.

However we *must* have tests verifying this stuff.

kominak:
I'm glad you found that useful. Feel free to modify it any way you like it, including adding some tests :) I should've added them myself.

I agree that calling `.toPromise(T)` on top of every "legacy" promise is a bit awkward, but we couldn't find any better (less weird) way either.

PR #395, opened 2016-03-29
Fix when dismissing a promiseViewController

### LINKED ISSUE #394: UIViewController promiseViewControlller does not always dismiss
I tracked the problem down, and it has to do with presenting a view controller from a view controller that can't handle the request, and so the view controller forwards it to its parent (as stated in the Apple docs for presentViewController:animated:completion.  The change in PromiseKit is to the finally block.  To be completely correct it should be changed from:

    promise.finally(^{
        //TODO can we be more specific?
        [self dismissViewControllerAnimated:animated completion:nil];
    });

to

    promise.finally(^{
        //TODO can we be more specific?
        [vc2present dismissViewControllerAnimated:animated completion:nil];
    });

or

    promise.finally(^{
        //TODO can we be more specific?
        [vc2present.presentingViewController dismissViewControllerAnimated:animated completion:nil];
    });

It is a pretty simple change, but I can create a pull request if you want.

Thanks, and let me know if you have any questions.

jgrantr:
The proper way to dismiss a the viewController is to call
dismissViewControllerAnimated:completion on
vc2present.presentingViewController

Fixes #394

### DIFF
--- a/Categories/UIKit/UIViewController+AnyPromise.m
+++ b/Categories/UIKit/UIViewController+AnyPromise.m
@@ -15,7 +15,7 @@
 
 - (AnyPromise *)promiseViewController:(UIViewController *)vc animated:(BOOL)animated completion:(void (^)(void))block
 {
-    id vc2present = vc;
+    UIViewController *vc2present = vc;
     AnyPromise *promise = nil;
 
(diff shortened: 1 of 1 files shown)


nathanhosselton:
Thanks again!

PR #396, opened 2016-03-30
Remove test targets to build together with framework targets

ghost:
Edited the build schemes for the iOS and OXS framework targets to not also build the test targets.  This gets rid of the following build warnings when build the frameworks via Carthage and Xcode 7.3:

Code Sign warning: CODE_SIGN_ENTITLEMENTS specified without specifying CODE_SIGN_IDENTITY. It is not possible to add entitlements to a binary without signing it.
warning: All interface orientations must be supported unless the app requires full screen.
warning: A launch storyboard or xib must be provided unless the app requires full screen.

The last two warnings  can easily also be fixed by setting the key `UIRequiresFullScreen` in the UI test target's `Info.plist`, however, for the code signing warnings, I could find no such easy fix other than just not build them together with the frameworks.  

I do not know what could break by not building the test targets together with the framework targets.  The discussion here mentions the approach I've taken: http://stackoverflow.com/questions/26109851/code-signing-is-required-for-product-type-unit-test-bundle-in-sdk-ios-8-0

### DIFF
--- a/PromiseKit.xcodeproj/xcshareddata/xcschemes/PMKOSX.xcscheme
+++ b/PromiseKit.xcodeproj/xcshareddata/xcschemes/PMKOSX.xcscheme
@@ -28,36 +28,6 @@
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
-         <TestableReference
-            skipped = "NO">
-            <BuildableReference
-               BuildableIdentifier = "primary"
-               BlueprintIdentifier = "6388E1A11AD315F90074D85B"
-               BuildableName = "PMKTests.xctest"
-               BlueprintName = "PMKTests"
-               ReferencedContainer = "container:PromiseKit.xcodeproj">
-            </BuildableReference>
-         </TestableReference>
-         <TestableReference
-            skipped = "NO">
-            <BuildableReference
-               BuildableIdentifier = "primary"
-               BlueprintIdentifier = "63533B321AE57A9300C10972"
-               BuildableName = "PMKAPlusTests.xctest"
-               BlueprintName = "PMKAPlusTests"
-               ReferencedContainer = "container:PromiseKit.xcodeproj">
-            </BuildableReference>
-         </TestableReference>
-         <TestableReference
-            skipped = "NO">
-            <BuildableReference
-               BuildableIdentifier = "primary"
-               BlueprintIdentifier = "63CA14481AE9FBD000223904"
-               BuildableName = "PMKOSXCategoryTests.xctest"
-               BlueprintName = "PMKOSXCategoryTests"
-               ReferencedContainer = "container:PromiseKit.xcodeproj">
-            </BuildableReference>
-         </TestableReference>
       </Testables>
       <MacroExpansion>
          <BuildableReference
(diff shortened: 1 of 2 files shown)


nathanhosselton:
> I do not know what could break by not building the test targets together with the framework targets.

Nothing breaks to my knowledge, but this does prevent us from conveniently running all tests associated with a target. Instead, they much be run individually or associated with a new scheme that groups them together (e.g. "PMKiOSTests" which would link the test targets previously associated with the PMKiOS scheme).

The latter of these alternatives isn't terrible but it does feel unnatural to me and I think shouldn't be necessary. It's unfortunate that associated test targets must be built with their parent target even when tests are not being run, but Xcode doesn't appear to give us any way around that. Perhaps it is better to simply code sign the test targets? I'm not sure why they are not currently, and I can't think of what issue this could cause.

I'm also not sure why this warning is appearing in the first place. Taking it at face value, it should only appear when code signing entitlements are specified, but there are none for any of the test targets. Only the host app target for UI tests has code signing entitlements set, and it also has a code signing identity set.

If you can, try code signing the test targets to see if that removes the warning without adverse side effects. If so, I think that's probably the way to go.

tentenalex:
> should only appear when code signing entitlements are specified, but there are none for any of the test targets

The code signing entitlements are added to the `PMKiOSUITestsHost` target (file on disk is `Tests/Entitlements.plist` and, judging by the contents, appear to have been added to disable the authorization dialogs for automated UI testing(?).


> Perhaps it is better to simply code sign the test targets?

Since I'm getting the dependency via Carthage and the build warnings appear on `carthage update`, I'm not sure what I would have to do to influence the signing process to work different to the way it does now. Any suggestions?

DesignatedInitializer:
I've just merged my private account and my professional account, that's why I'm appearing as ghost above.

nathanhosselton:
> Since I'm getting the dependency via Carthage and the build warnings appear on carthage update, I'm not sure what I would have to do to influence the signing process to work different to the way it does now. Any suggestions?

Sorry, I realize now that's not a thing you can trivially test.

Thinking on this more, is this causing Carthage update to fail or otherwise preventing use of PromiseKit via Carthage? Or is it merely an informational warning that Carthage passes along from xcodebuild? It sounds like it's the latter, but I'm not sure why I'm not also seeing these warnings.

nathanhosselton:
OK, I assume xcodebuild doesn't set a code signing identity for targets by default. I have one set for PMKiOSUITestsHost on my end, so I do not see the warnings. But when using `carthage update`, it builds before a valid identity can be set. Thus, the warnings appear.

Not really sure what can be done about this. I thought it was that the test targets needed a code signing identity set simply because they were associated with the test host app target which contains entitlements. But it sounds like the problem is simply what I stated above, which means there isn't anything that can be done as I would expect that even creating a new scheme specifically for the tests would still cause this issue. Sorry if you understood this from the start and had to wait until now for it to click with me.

Your PR does fix the issue, but it also prevents us from being able to run the tests, so I cannot merge it. Assuming I'm understanding the problem correctly (and perhaps I am not), this seems like an issue that Carthage needs to solve rather than developers working around it.

DesignatedInitializer:
Sorry for taking so long to reply.  The Carthage build isn't failing, but it's printing a number of warning and we'd just like it to build without any warnings.  Understand what you're saying, will close the PR.

DesignatedInitializer:
Looks like I don't have access rights to close the PR as I created it with a now deleted account, if you can, please go ahead and close it.

PR #397, opened 2016-03-31
Provide a way out in case Swift compiler really doesn't know which method or property is meant by "error"

### LINKED ISSUE #347: 3.0.0: Cannot call value of non-function type 'ErrorType?'
Installed PromiseKit 3.0.0 via Cocoapods.  The following code generates a compile error (Xcode Version 7.2 (7C68)):

```
let (p2, _, _) = Promise<Void>.pendingPromise()
p2.error { error -> Void in
        NSLog("Error: %@", error)
}
```

This appears to be because `Promise+Properties.swift` defines `public var error: ErrorType?` and `Promise.swift` defines `public func error(policy policy: ErrorPolicy = .AllErrorsExceptCancellation, _ body: (ErrorType) -> Void)`.

I'm not sure Swift allows you to have properties with the same names as functions.  The Swift compiler is getting confused.

ReneB:
Fixes #347

### DIFF
--- a/Sources/Promise+Properties.swift
+++ b/Sources/Promise+Properties.swift
@@ -14,6 +14,14 @@ extension Promise {
     }
 
     /**
+     - Returns: The error with which this promise was rejected; `nil` if this promise is not rejected.
+       Alternative for the cases where the Swift compiler gets confused over the name of the `error` property
+    */
+    public var errorValue: ErrorType? {
+        return self.error
+    }
+
+    /**
      - Returns: `true` if the promise has not yet resolved.
     */
     public var pending: Bool {
--- a/Sources/Promise.swift
+++ b/Sources/Promise.swift
@@ -391,6 +391,14 @@ public class Promise<T> {
     }
 
     /**
+     Provides an alias for the `error` function for cases where the Swift compiler gets into trouble inferring the right
+     types. If `error` does not work, consider using `onError` instead which does the same thing.
+    */
+    public func onError(policy policy: ErrorPolicy = .AllErrorsExceptCancellation, _ body: (ErrorType) -> Void) {
+        error(policy: policy, body)
+    }
+
+    /**
      The provided closure is executed when this promise is rejected giving you
      an opportunity to recover from the error and continue the promise chain.
     */
--- a/Tests/CorePromise/ErrorUnhandler.test.swift
+++ b/Tests/CorePromise/ErrorUnhandler.test.swift
@@ -116,6 +116,20 @@ class ErrorHandlingTests_Swift: XCTestCase {
         }
     }
 
+
+    // an alias called `onError` exists for the `error` function
+    func test7() {
+        twice { promise, ex in
+            PMKUnhandledErrorHandler = { err in
+                XCTFail()
+            }
+
+            promise.onError { error in
+                ex.fulfill()
+            }
+        }
+    }
+
     func testDoubleRejectDoesNotTriggerUnhandler() {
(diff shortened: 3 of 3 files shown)


nathanhosselton:
Looks good to me 👍. Will get merged tomorrow and pushed up to cocoapods if I don't hear otherwise from @mxcl.

nathanhosselton:
Hey sorry for the delay. Thought of something in the meantime though. What expectation are we setting with these alternatives in terms of their longevity? Obviously they're meant to be temporary, so should we explicitly state that in their documentation, or even mark them as deprecated from the start to more strongly convey that? Or is that to heavy handed given that we can't necessarily expect this issue to be resolved even by Swift 3.0?

ReneB:
If you ask me, any decision will do. I think these could be removed as soon as the Swift compiler stops bugging out on us. Having said that, as I said in #347 (the thread that inspired this pull request), I think having the `error` property and an `error()` function on the same object might not be valid Swift, in which case we might want to look for a slightly more permanent solution. 

The Promises/A+ spec has some terminology for what Promises are expected to do (things like `onFulfilled` and `onRejected`). It might be smart to let that terminology inspire any naming decisions.

nathanhosselton:
Ok, I think these aliases should be marked as deprecated then since they're only intended to be used as a temporary workaround for cases when the Swift compiler cannot be made to distinguish between the originals, and they will be removed as soon as...

1. Swift is updated to better handle our names, or 
2. we decide on permanent name alternatives...

whichever comes first. The deprecation warnings and/or documentation should communicate this so users don't become dependent on these aliases or use them inadvertently. It think this PR can be merged in after that and we can start planning what the final solution will be.

Naming is hard. PK's current prefix-less naming scheme felt more complete when we could still use `catch`. Though I like the name `onError` better than simply `error`, it's inconsistent with our other handler names, so ideally the naming scheme as a whole would be revised before we keep it (which is an option of course, though I like all of our other names). Otherwise we need another name for `error()`, since I like `error` as the property name.

I digress, though; that's a conversation for a different thread.

@ReneB do you want to include the deprecation warnings in your PR, or do you prefer I add them? Either way is fine with me.

ReneB:
I'm currently really pressed for time, so if you could do it, that would be great. Otherwise, I'll get around to it somewhere around next Tuesday or so, which is also fine with me :-)

nathanhosselton:
@ReneB no problem. Will get this added in and merged by tomorrow probably.

nathanhosselton:
@ReneB Done. Thank you again for the excellent recommendation and the PR!

gsabran:
I'm having issues with return types:

```swift
private func connect() -> Promise<Void> {
    return tryToConnect()
        .then { payload -> Void in
            // do something here
        }
        .onError { _ -> Void in
    }
}
```
Raises the error `Cannot convert return expression of type '()' to return type 'Promise<Void>' (aka 'Promise<()>')`

`tryToConnect` is of type `() -> Promise<JSON>` if that helps

ReneB:
First off, this kind of things is often slightly dependent on the Swift compiler version you're using. Are you on XCode 7 or 8 (or some other compiler)?

mxcl:
```
private func connect() -> Promise<Void> {
    return tryToConnect()
        .then { payload -> Void in
            // do something here
            // don't return anything
        }
        .onError { _ -> Void in
    }
}
```

This is because `onError` returns `Void`, not `Promise<Void>`.

gsabran:
duuh sorry about that!

PR #401, opened 2016-04-06
Add xcuserdata to .gitignore

ultramiraculous:
When using PromiseKit as a subproject, Xcode keeps adding xcuserdata folders. If you're using git submodules it means that the submodule is perpetually considered out of sync while you're developing.


nathanhosselton:
Thanks! Been meaning to add this myself and kept forgetting.

Issue #280, opened 2015-09-25
Add ability to use custom queue for `report`


ldiqual:
In Xcodo 7's UI Tests, the main queue is blocked when calling `waitForExpectationsWithTimeout(timeout:handler:)`, which means you can't call a Promise and use the default queue in the `report` block. It would be perfect if we had the ability to call the report block in another queue than the main one.

mxcl:
PromiseKit's tests use report and expectations all over the place.

Can you provide a test project that verifies your bug? Thank you.

ldiqual:
@mxcl The problem only occurs in UI test targets when `waitForExpectationsWithTimeout` is called right after a `then/catch` is chained on a promise. The `then` block is not blocked when performed on another queue than the main queue. Alamofire has the same issue with its default queue here: https://github.com/Alamofire/Alamofire/issues/809. I'll provide a test project asap, thanks for asking.

Side note: I tried to checkout the project and run the tests but I'm unable to setup the carthage dependencies (it keeps asking me for my github credentials when I run `carthage bootstrap`). How do you usually proceed to checkout and run your tests?

ldiqual:
@mxcl Here's the example project: https://www.dropbox.com/s/2c6vfqgxtuzltlj/PromiseKitReportQueueExample.zip?dl=0

The interesting stuff is in PromiseKitReportQueueExampleUITests.swift, you'll see 3 tests as part of the UI test target:
 - A successful promise with a `then` on the default queue (fail)
 - A successful promise with a `then` on a concurrent queue (pass)
 - An unsuccessful promise with a `report` on the default queue (fail)

I couldn't find any documentation on the behavior of `waitForExpectationsWithTimeout` in a UI test target, but I believe that it blocks the main queue from receiving further messages until all expectations are met. It doesn't block the actual UI though (you can still interact with the app in the simulator), which is weird.

ldiqual:
@mxcl More info on the `waitForExpectationsWithTimeout` behavior: http://masilotti.com/xctest-documentation/Classes/XCTestCase.html#//api/name/waitForExpectationsWithTimeout:handler:

> -waitForExpectationsWithTimeout:handler: runs the run loop while handling events until all expectations are fulfilled or the timeout is reached. Clients should not manipulate the run loop while using this API.

mxcl:
We don’t manipulate the run loop. We just `dispatch_async` onto it. This seems weird to me, Xcode UI tests can't work this way, so much stuff would break.

[...]

PR #405, opened 2016-04-14
Add ability to change the default dispatch queue for all promises

programmerdave:
This PR allows you to change the default dispatch queue for all promises in Swift and in ObjC.

It fixes: #280 and allows you to use Zalgo in unit tests so that you don't have to depend on dispatch_async so you can get fast deterministic tests (http://martinfowler.com/articles/nonDeterminism.html)


mxcl:
I like it.

1. I think there should be just one function, do it in objc so it is accessible from both
2. I think it should be named `PMKDefaultDispatchQueue()` since that is more consistent with other such things in PromiseKit.

programmerdave:
@mxcl I changed the name, made it accessible from both, and all tests pass (aside from an existing Issue which I submitted a separate PR for). Thanks!

nathanhosselton:
Looks good to me.

mxcl:
LGTM! Many thanks.

programmerdave:
No problem! Thanks! Do you know when this would be merged into master?

nathanhosselton:
Tomorrow morning if not tonight!

nathanhosselton:
Squashed and merged. Closing manually.

Thanks again, @programmerdave!

PR #406, opened 2016-04-15
Make Image Picker test more reliable

programmerdave:
Due to timezone differences, I think my simulator is showing different times for the photos so the test fails. I made it use a predicate to find the right photo.


nathanhosselton:
Nice. Thank you. This test is a little frustrating because the dates also seem to inconsistently change with new versions of iOS. So I'm always updating it. Hopefully this will make it more resilient in addition to solving the time zone issue you found.

programmerdave:
If you want to make it more resilient in the future, and all you need is any photo, you can just use the ```"SELF.label LIKE %@", "Photo"``` predicate instead so it's even more future proof

PR #425, opened 2016-05-15
Add Swift 3 swift-DEVELOPMENT-SNAPSHOT-2016-05-09 compatibility.

TheArtOfEngineering:
Added basic support for Swift 3 and the Swift Package manager, just to get things started. Did not add test compatibility due to complications with dependencies.

This is not backwards compatible and **should be merged into a branch other than master**. I'm just putting it on the radar. Due to the sweeping syntax changes with Swift 3 backwards compatibility is not maintainable (a change to a function declaration requires a duplication of the function). 

There were some newly annotated nullable APIs. Theses were force unwrapped where they occurred.

Enums cases in Swift 3 now have lowerCamelCase names, so I converted the PromiseKit enums to conform to this.

The intention here was also to preserve PromiseKit's current public API.


mxcl:
It appears this effort was duplicated. Sorry.

TheArtOfEngineering:
Understandable, thanks anyway!

PR #426, opened 2016-05-17
Add ability to use custom queue for error #280 #405

yannickl:
Hi there,

I really need to choose the queue where the error runs because I provide a library that uses PromiseKit and I don't want update the `PMKDefaultDispatchQueue` for the entire project.

I think the `error(on: queue)` can be pertinent in some use cases.

Best,

Yannick

### DIFF
--- a/Sources/Promise.swift
+++ b/Sources/Promise.swift
@@ -358,14 +358,33 @@ public class Promise<T> {
      of a chain. Often utility promises will not have a catch, instead
      delegating the error handling to the caller.
 
-     The provided closure always runs on the main queue.
+     The provided closure runs on PMKDefaultDispatchQueue by default.
 
      - Parameter policy: The default policy does not execute your handler for cancellation errors. See registerCancellationError for more documentation.
      - Parameter body: The handler to execute if this promise is rejected.
-     - SeeAlso: `registerCancellationError`
+     - SeeAlso: `errorOnQueue`
     */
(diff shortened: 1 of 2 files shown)


mxcl:
You didn't say why you need this.

yannickl:
Sorry, here is a link to an example (this is more explicit): https://github.com/yannickl/AwaitKit/blob/master/Sources/AwaitKit.swift#L46

My library uses your PromiseKit as promise implementation. As you can see I need to block the main thread with a semaphore. So if an error occurred, I need to notify the semaphore.

I could update the `PMKDefaultDispatchQueue` but if the user also modify it on his own side (with the main queue), it can alter my library behaviour and entered in a deadlock.

mxcl:
K, this seems like a good enough use-case to me. There have been good enough uses in the past, the problem is Swift+Xcode has hated error having another parameter and has been busted because of it.

I suggest we roll this into PMK4 / Swift3, if you can wait that long? I guess not.

In the meantime I'm cool with an additional method like:

    errorOnQueue(…)

Ugly, but I really don't want to break compilation for Swift 2.x users (it breaks the parser, foo foo foo).

yannickl:
Yes, of course, I would like to have this feature before the release of PMK4 to be able to release my library.

I'll send you a pull request for with `errorOnQueue()` method to not break the compilation for Swift 2.x users.

Thank you!

yannickl:
I just updated the pull request. I hope it'll be ok for you.

mxcl:
LGTM thanks!

PR #432, opened 2016-05-25
Removed OMGHTTPURLRQ dependency

leoMehlig:
Removes the OMGHTTPURLRQ carthage dependency since it is not used in the project build by carthage. If the user uses the relative category he/she should add the dependency to the cartfile manually. This also reduces carthage build time and avoids confusion.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,0 @@
-github "mxcl/OMGHTTPURLRQ" ~> 3.1.2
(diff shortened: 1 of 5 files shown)


mxcl:
I agree in principle, I worry people won't realize they need it, but ultimately agree Carthage users are the kinds of users to read the docs. LGTM.

leoMehlig:
I've added a small note above all import statements. This should probably point everyone in the right direction.

PR #433, opened 2016-05-26
Remove watchos exclusion of NSURLSession extension

lammertw:
I've read in #265 that watchos support was added but that it wouldn't be necessary to have support for network requests. I'm not sure why that was concluded since it's perfectly possible to perform network requests from a watch app. The function used from the NSURLConnection Promise extension does not work but the one from the NSURLSession Promise extension works fine. I've tested it from a watch app.

### DIFF
--- a/PromiseKit.podspec
+++ b/PromiseKit.podspec
@@ -90,7 +90,7 @@ Pod::Spec.new do |s|
   s.subspec 'Foundation' do |ss|
     ss.ios.source_files = Dir['Categories/Foundation/*'] - Dir['Categories/Foundation/NSTask*']
     ss.osx.source_files = 'Categories/Foundation/*'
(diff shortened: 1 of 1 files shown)


mxcl:
WatchOS 1.x didn't support such things, WatchOS 2.x does.

lammertw:
But watchos support was only added for watchos 2 right? (According to #265)

mxcl:
The history of this stuff is fuzzy for me. Anyway if it works, it's great! and much thanks.

lammertw:
You're welcome!

lammertw:
This changed has been undone in commit eca0eba82db2dcecc75c7eb648a5fefe81e7b672 and therefore the NSURLSession+Promise.swift is again unavailable to watchos targets.

PR #439, opened 2016-06-01
join for void promises

jacobwallstrom:
Without this fix I get a compilation error

No 'join' candidates produce the expected contextual result type
'Promise<Void>' (aka 'Promise<()>')

when joining void promises.

### DIFF
--- a/Sources/join.swift
+++ b/Sources/join.swift
@@ -20,6 +20,10 @@ public func join<T>(promises: Promise<T>...) -> Promise<[T]> {
     return join(promises)
 }
 
+public func join(promises: [Promise<Void>]) -> Promise<Void> {
+    return join(promises).then { (result:[Void]) in return Promise() }
+}
+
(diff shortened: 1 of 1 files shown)


mxcl:
Just out of interest, Swift version? Xcode version?

Swift should handle this, I think, but since it doesn't, LGTM.

jacobwallstrom:
I'm using Xcode 7.3 (7D175).

When thinking some more about this I realize that the calling site is expecting Promise<Void> but the join is returning Promise<[Void]>. So a workaround is to at the call site do `join(promises).then { _ -> Void in }`

So this could be seen as a convenience for void promises since an array of voids are equivalent to just void.

mxcl:
Ahh ok, yes I was expecting swift to handle the `[Void]` case, but indeed I acknowledge that what you really want is a `Void` one, so indeed, **provided** this doesn't make join more ambiguous to use, LGTM.

I worry about the ambiguity when trying to write simple chains.

PR #445, opened 2016-06-12
Promise Queue: The first approach.

shergin:


nathanhosselton:
Hey there. Thanks for this! Currently reviewing and will provide feedback (by tomorrow I hope). I know you mentioned that you will be including tests. We will need them before merging, so feel free to push those whenever you have them. Things like ensuring the max pending promises and max promises in queue are not exceeded and checking for races on their counters come to mind immediately.

shergin:
Hey Nathan,
This is actually my production code.
I will definitely add tests after we will conceptually agree about entire API. I just don't want to rewrite these tests. )

Thank you!

nathanhosselton:
That's fair. This looks good on first glance and I don't expect major changes but better safe than sorry with your time, for sure.

nathanhosselton:
Alright so after spending a bit of time reviewing and playing with it, I don't have any concerns with the API; I think it's solid. But I am seeing some opportunity for refinement. Potentially. I'm going to take a bit more time and construct some proposals to see what you think. I think this could be elevated above a niche feature to something a bit more intrinsic to PK itself - perhaps even integrated into existing functions (e.g. `when`, `join`). But, again, I need to think on it a bit more. It may end up being better as is.

shergin:
That's would be awesome! I will add the tests soon... I hope.

nathanhosselton:
Interesting. I've not see the `GeneratorType` protocol used before. Feel free to push that change up if you'd like.

Sorry I have not followed up yet. I wanted to collect more concrete thoughts but while I'm here I'll go ahead and mention some of my observations thus far:

• @mxcl mentioned using a term other than `queue` and I feel like `chain` is a logical replacement.
• I disliked `generator` as a parameter name initially, but perhaps it will make more sense with the changes from your prior comment.
• Mirroring a comment from @mxcl, I feel the generator closure on `when` is unintuitive. Reading a call to the function, it isn't obvious what the trailing closure is for. This is somewhat exacerbated by the fact that our API frequently uses trailing closures as completions. A possible solution may be to require passing a fully constructed `PromiseQueue` object to the function, but I also liked the abbreviated nature of it preventing this extra effort, so I'm conflicted. Since the function doesn't actually take any promises, perhaps we shouldn't even call it `when`.
• I'll handle all the documentation so don't worry about this.

I'll of course keep digging through and thinking on all this but if you have thoughts please feel free to share.

Sources/PromiseQueue.swift
```diff
@@ -0,0 +1,138 @@
...
+        catch (let error) {
+            self.pendingPromises -= 1
+            item.reject(error)
+            self.dequeue()
+            return
+        }
+
+        promise
+            .always {
+                self.pendingPromises -= 1
```
mxcl: Use `zalgo`, it's safe here and we should use it with library code.

mxcl:
I wonder about the maxQueue parameter (again), the example given was for `concurrently: 1, maxQueue: 1`, are there other situations where maxQueue would be something other than 1? If not then I don't think we should have the parameter in general, and instead have a separate function of the `1` case. The reason I think this way is the `error` that may be thrown. I strongly dislike functionality that may error at runtime and which you cannot reason about whether or not it will error at runtime. The maxQueue parameter may error at runtime and you cannot be sure if it will or not.

Thinking about, eg. a save queue, you want at most TWO promises in the queue, the one executing and the one pending. Thus if a third is added you would want the second to be discarded and the third to be run instead once the first is finished.

So well, indeed, we don't have a good answer here yet.

I apologise. I know how much it sucks to write code you want in a library and have it endlessly debated, the thing is: I feel PromiseKit is in a pretty good state right now, and I only want to add features to it that are pretty good. So we are being cautious.

Libraries should guarantee quality, in my opinion. And it's so important that they don't add features once their users trust them, and that those features are not carefully considered.

shergin:
Nathan, Max, you are awesome.
I love to have discussion with you about one of my favorite themes. )

I absolutely agree that the library should be solid and just right amount of functionality. Or be more modular, so we can have additional module like "Operations" witch can contain set of complicated functionality like queues, barriers, and so on.

I have implemented another approach in PR #449, so: 
 * This implementation is much more simpler and I believe more fit to existing PromiseKit API.
 * It is using generators instead of blocks which I found more conceptual and swift-idiomatic.
 * I have tried to use `zalgo` but... to be honest I don't understand this (`zalgo`-`waldo`) concept very well yet.
 * I will add more tests little bit later.
 * I get fid of `maxQueue` parameter because it doesn't fit to new approach at all.

shergin:
Guys,
What do you think about the second (#449) approach?

mxcl:
Closing in lieu of #449.

Issue #444, opened 2016-06-07
Promise-based queue with limited number of "workers"


shergin:
I need something like npm-library `promise-queue` (https://www.npmjs.com/package/promise-queue) but in Swift and for PromiseKit.
So, is there any chance that PromiseKit will contain something similar?

mxcl:
Synopsis of what you want? Link doesn't really explain itself.

shergin:
Let's assume that we have a number of items (or stream of items) that we have to process. And we have promise-fabric that can process this type of items.
Usually we can use `when()` function, like this
```
func processor(item: Item) -> Promise<ProcessedItem> {
    return Promise { ... }
}

let items: [Item] = [....]

when(items.map { processor($0) }).then { processedItems in
    // Done!
}
```

That really great but... I know that I have system resources that allow me process only 3 items in parallel (otherwise iOS will kill my app under memory pressure).

So I decided to use pattern that very common (kind of) in front-end-promise-world: promise queue.

It can be done like this:

```
func processor(item: Item) -> Promise<ProcessedItem> {
    return Promise { ... }
}

let items: [Item] = [....]

let queue = PromiseQueue<Item, ProcessedItem>(maxConcurrent: 5, maxQueue: 100)
for item in items {
    queue.add(processor(item))
}

queue.resultPromise.then { processedItems in
   // Done!!
}
```

Disclaimer: I am aware of how GCD's queues work, I believe it cannot be done thought this kind of queues.

My question actually is: Dear mxcl, as maitainter, do you think that such kind of (more high-level) functionality has to be in (pretty low-level, which I really love) library like PromiseKit or not?

PS: This pattern also is very useful with UI/UX programming when we need to deal with some user input but we can throttle some of pieces of user input.

mxcl:
Makes sense. We should have this.

mxcl:
Seems like some parameters to when may be logical? Or would this just be non-intuitive to people experienced with Promises?

```swift
when(promises, concurrently: 5).then {
    //…
}
```

The maxQueue parameter seems a little esoteric.

shergin:
I believe, we actually cannot use/extend `when` because it accepts promises, so if we already have the promises, it immediately means that they already started resolving process (but we want to postpone it).

`maxQueue` parameter can be useful in UI programming when we are implementing something like asynchronous throttling. I have used it.
[...]

PR #449, opened 2016-06-21
`when` with limited concurrently pending promises.

shergin:
The second approach for #444.


Sources/when.swift
```diff
@@ -86,3 +86,88 @@ public func when<U, V>(pu: Promise<U>, _ pv: Promise<V>) -> Promise<(U, V)> {
...
+    var generator = promiseGenerator
+
+    return Promise { fulfill, reject in
+        var pendingPromises = 0
+        var values: [T?] = []
+        var rejected = false
+        var counter = 0
+
+        func extendArraySize(size: Int) {
+            if size > values.capacity {
```
shergin: I just realized that probably I should not do it because standard library should extend capacity automatically.

mxcl:
Looks great. Nice use of `AnyGenerator`. I approve of using the stdlib.

shergin:
Cool!
So, I will add some more tests and we good to go (merge)?

nathanhosselton:
@shergin I would say so.

shergin:
I have just added several tests for it.
What do you think?

mxcl:
Great, appreciated :)

The documentation for this function is a copy and paste from the other when. Can we make it describe what is different?

Running all PMKTests crashes for me:

screen shot 2016-07-16 at 11 13 47 am

Also, this is not thread safe.

shergin:
Ough! Thanks! But I thought it is not finished yet. ) I am on vacation right now so I have not time to work on it. Sorry.
I have a fix for the crash that was shown on screenshot and I would love to discuss about thread-safety. I will open new PR for that. )

PS:
Max, nice avatar!

mxcl:
I fixed the issues in subsequent commits, though… you should review!

shergin:
@mxcl Awesome! Your approach is much better! Thank you, Max!

nobre84:
Hey! This is a great under documented feature! Any reason why it isn't available for when(resolved:) as well ?

mxcl:
No reason, PR welcome!

Issue #450, opened 2016-06-21
CLLocationManager extension can call reject multiple times


tcunning:
1. In the iOS-9 simulator don't set a default location 
2. Invoke CLLocationManager.promise(.WhenInUse).then { ... }
3. Allow location services
4. The locationManager delegate didFailWithError is getting called twice by location manager.   This is causing the PromiseKit reject method (see CLLocationManager+Promise.swift line 64) to get called twice which cases the following warning:

PromiseKit: Warning: reject called on already rejected Promise: Error Domain=kCLErrorDomain Code=0 "(null)"

tcunning:
Looks like this is resolved in the latest

tcunning:
This issue only effects the swift implementation as the objective-c implementation shuts down the location manager after an error occurs.

mxcl:
Merged: https://github.com/mxcl/PromiseKit/commit/633fa83ca0c309dc54837db889fab820db5f27a3

PR #452, opened 2016-06-21
Fix CLLocationManager extension can call reject multiple times

tcunning:
manager.promise.always was insufficient as a means to stop listening for location manager events.  It was possible to receive multiple events from the location manager that could cause fulfill or reject to get called multiple times.  This is fixed by having the location manager delegates perform their own cleanup.  This guarantees fulfill or reject will only get called once. #450

Used a similar solution to what was used for the Objective-C implementation.


mxcl:
@nathanhosselton please merge

PR #458, opened 2016-06-28
Added LICENSE file and removed '@see README' from podspec

tomquist:
When using the acknowledgement.plist, generated by CocoaPods, it only contains the text '@see README' for PromiseKit. Adding a separate LICENSE file will enable CocoaPods to fetch the license text and include it within the plist.

### DIFF
--- a/LICENSE.markdown
+++ b/LICENSE.markdown
@@ -1,0 +1,19 @@
+Copyright 2015, Max Howell; <mxcl@me.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


mxcl:
Well. I deliberately didn't want a separate license file. It's untidy.

tomquist:
Ok, then there are several alternatives:

1. Remove license section from README to avoid duplicate license information
2. Add full license text to podspec, keep license section in README and delete LICENSE file

I'd be happy to change my PR accordingly. I think being able to use the acknowledgement Plugin for CocoaPods is a very important.

mxcl:
Put a link to the README in the license section of the podspec.

PR #466, opened 2016-07-09
Fix compile error in when.m

sryze:
The `PMKProgress` struct was missing a `fractionCompleted` member which resulted in a compile error if `PMKDisableProgress` was defined.


mxcl:
@nathanhosselton please merge

PR #485, opened 2016-08-03
Update README to reflect new documentation URLs

grantamos:
The current documentation URLs are broken due to recent gh-pages updates.  I've updated the links to reflect the proper github page URLs.


mxcl:
Thanks, these should have been redirecting so I have also fixed the redirects:

http://promisekit.org/introduction/

PR #491, opened 2016-08-07
make AnimationOptions members public

allen-zeng:
## What is this?
This change opens up the `AnimationOptions` members `None`, `Appear` and `Disappear` as public.

## Why?
`promiseViewController(_:animate:fulfills:completion:)` asks for `AnimationOptions` as a parameter, however none of its members are marked `public` so users of the framework can't actually get to them. The only alternative is by calling the public initialiser `init(rawValue:)`, which is rather fragile.

### DIFF
--- a/Categories/UIKit/UIViewController+Promise.swift
+++ b/Categories/UIKit/UIViewController+Promise.swift
@@ -40,9 +40,9 @@ extension UIViewController {
         }
         public let rawValue: Int
 
-        static let None      = AnimationOptions(rawValue: 0)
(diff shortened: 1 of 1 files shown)


allen-zeng:
Hmm ok, the failed check seemed to have stalled, is there a way to rerun it?

mxcl:
My bad, this is somewhat the problem with our tests versus how PMK is built for CocoaPods. With CocoaPods all members are put in PromiseKit.framework, but we don't do that in the Xcode project as that would mean all categories would be imported for all uses of the framework which is not efficient for most people.

PR #497, opened 2016-08-10
implement fulfillment type

allen-zeng:
`FulfillmentType` is now used within `promiseViewController`.

### DIFF
--- a/Categories/UIKit/UIViewController+Promise.swift
+++ b/Categories/UIKit/UIViewController+Promise.swift
@@ -39,7 +39,7 @@ extension UIViewController {
         return promiseViewController(vc, animate: [.Appear, .Disappear], completion: completion)
     }
 
-    public func promiseViewController<T>(vc: UIViewController, animate animationOptions: PMKAnimationOptions = [.Appear, .Disappear], fulfillmentType: FulfillmentType = .OnceDisappeared, completion: (() -> Void)? = nil) -> Promise<T> {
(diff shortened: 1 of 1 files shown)


Categories/UIKit/UIViewController+Promise.swift
```diff
@@ -39,7 +39,7 @@ extension UIViewController {
         return promiseViewController(vc, animate: [.Appear, .Disappear], completion: completion)
     }
 
-    public func promiseViewController<T>(vc: UIViewController, animate animationOptions: PMKAnimationOptions = [.Appear, .Disappear], fulfills: FulfillmentType = .OnceDisappeared, completion: (() -> Void)? = nil) -> Promise<T> {
+    public func promiseViewController<T>(vc: UIViewController, animate animationOptions: PMKAnimationOptions = [.Appear, .Disappear], fulfillmentType: FulfillmentType = .OnceDisappeared, completion: (() -> Void)? = nil) -> Promise<T> {
```
nathanhosselton: Why the method signature change?

nathanhosselton:
Otherwise this LGTM.

Categories/UIKit/UIViewController+Promise.swift
mxcl: Yes, please don't change the name, strictly this is a major semantic version bump.

allen-zeng: Good point. I've added the label back in. The parameter name is still `fulfillmentType`, so that it reads more nicely when used.

allen-zeng:
I'm assuming we are following semantic versioning? I do wonder whether this constitutes a minor version bump though. The behaviour _has_ changed, if only subtly.

PR #498, opened 2016-08-11
Let `animated` and `completion` do something

SilverEnderman:
Not sure why you're hard-coding values where the default parameters are more than sufficient. Why not let the caller decide whether he wants presentViewController() to animate, and/or do something when it's finished? Is this intentional? If so, the parameters should be removed entirely from the function definition, to prevent confusion.

### DIFF
--- a/Categories/UIKit/PMKAlertController.swift
+++ b/Categories/UIKit/PMKAlertController.swift
@@ -71,7 +71,7 @@ public class PMKAlertController {
 extension UIViewController {
     public func promiseViewController(vc: PMKAlertController, animated: Bool = true, completion: (() -> Void)? = nil) -> Promise<UIAlertAction> {
         vc.retainCycle = vc
-        presentViewController(vc.UIAlertController, animated: true, completion: nil)
+        presentViewController(vc.UIAlertController, animated: animated, completion: completion)
         vc.promise.always { _ -> Void in
             vc.retainCycle = nil
         }


mxcl:
Thanks for noticing and the fix. This was a mistake and not a deliberate decision.

SilverEnderman:
By the way, would you happen to know why the compiler doesn't see the category symbols? I only found out about this thing from GitHub, and digging through the source folders on my end... Xcode balks when I try to use any of the `promiseViewController` methods, or the `PMKAlertController`, claiming neither exist. I'm trying it via CocoaPods to see if it makes a difference, but would you happen to have any advice in the meantime?

mxcl:
With Carthage none of the categories are installed because there is no way to choose subspecs like there is with CocoaPods.

This used to be documented but apparently has disappeared in the recent docs rewrite.

You have to manually add the category sources to your project with Carthage.

mxcl:
Actually it is documented here: http://promisekit.org/docs/cocoa-promise/

SparrowBlaze:
Not using Carthage. Sorry, I should've specified. I actually downloaded the .xcodeproj and added the framework as an embedded binary.

Using CocoaPods now, and still getting similar troubles. Told it both PromiseKit (3.4) and PromiseKit/UIKit. Installed, updated, `import PromiseKit`, clean built, the works, and Xcode still doesn't know what the heck `promiseViewController()` is supposed to do.

mxcl:
Go into the Pods project and check that `UIViewController+Promise.swift` is there.

SparrowBlaze:
screen shot 2016-08-11 at 11 26 49 am
This the one?

SparrowBlaze:
The Swift header doesn't mention UIViewController at all, but that's probably more of the symptom than the cause...

mxcl:
The Swift header won't mention it since `Promise<T>` cannot be represented in ObjC.

Anyway, the file is there, so something is wrong with your build. Since you added the xcodeproj manually and then added it with CocoaPods it is possible the old PromiseKit is still around somehow and that is being imported, and the old one had no categories.

SparrowBlaze:
I deleted old PromiseKit before starting on CocoaPods. Same behavior. Rebuilding now, then I'll try deleting the DerivedData entry and rebuilding again.

SparrowBlaze:
Hmm... The build just finished, and I'm getting some strange symbol errors in the console... Gonna try tweaking build orders. Maybe it's not copying the sources before linking, or something...

*i have no idea what im doing*

PR #513, opened 2016-08-19
Updated Documentation

CalQL8ed-K-OS:
This mention in the documentation sounds like a carry over from the Obj-C implementation.
The remaining documentation when you remove this feels sufficient, as the code is sufficiently self documenting IMHO.
```
If the task fails the promise is rejected with code `PMKTaskError`, and
`userInfo` keys `PMKTaskErrorStandardOutputKey`,
`PMKTaskErrorStandardErrorKey` and `PMKTaskErrorExitStatusKey`.
```


mxcl:
Thanks. 👍🏻

PR #514, opened 2016-08-21
Fix podspec

lammertw:
There was an error in the Podspec, resulting in the following error when installing:

[!] Unable to find a specification for 'PromiseKit'.

[!] Unable to load a podspec from `PromiseKit.podspec`, skipping:

Pod::DSLError


mxcl:
Thanks and apologies.

PR #519, opened 2016-08-27
Fixes NSURLSessionDataTask never resolving its promise

wokalski:
A little fix. (unless I'm doing something terribly wrong)


mxcl:
Ahh. This is why we need tests for this part.

mxcl:
Pushed 3.5.1

PR #569, opened 2016-10-24
Make Promise public instead of open

zlangley:
Not sure if this is intentional. Does it make sense to subclass `Promise`? Are people doing this?


zlangley:
Ah, okay. Apparently allowing subclassing is intentional.

mxcl:
We use it in a number of places, but I'm not sure if it really justifies it being `open`. Could do the same tricks with a protocol, and this would probably be better anyway.

PR #570, opened 2016-10-24
Move contain_zalgo to DispatchQueue

zlangley:
I also wonder if something like `asyncIfNotZalgo` is a better name than `containZalgo`, since it wasn't obvious to me at first than `containZalgo` dispatched.


zlangley:
Actually I'm not sure about this. It's awkward to have an "instance" method that checks against the instance to determine what to do.

mxcl:
We can rename the method or document it more.

Probably `static func DispatchQueue.pmkAsync` would be enough, to indicate it is special to us.

PR #571, opened 2016-10-25
Document all public methods

zlangley:
Probably need some help on the `PMKJoint` documentation; I pretty much just stole the comment out of the message of the commit that added it. I tried adding a test to understand how it would be used, but I'm still not sure I get it. Some sample code would be useful!

### DIFF
--- a/03_JointTests.swift
+++ b/03_JointTests.swift
@@ -1,0 +1,53 @@
+import PromiseKit
+import XCTest
+
+class JointTests: XCTestCase {
+    func testJointPiping() {
+        let (promise, joint) = Promise<Int>.joint()
+
+        XCTAssert(promise.isPending)
+
+        let foo = Promise(value: 3)
+        foo.join(joint)
+
+        XCTAssertEqual(3, promise.value)
+    }
+
+    func testJointPipingPending() {
+        let (promise, joint) = Promise<Int>.joint()
+
+        XCTAssert(promise.isPending)
+
+        let (foo, fulfillFoo, _) = Promise<Int>.pending()
+        foo.join(joint)
+
+        fulfillFoo(3)
+
+        XCTAssertEqual(3, promise.value)
+    }
+
+    func testJointCallback() {
+        let ex = expectation(description: "")
+
+        let (promise, joint) = Promise<Void>.joint()
+        promise.then { ex.fulfill() }
+
+        Promise(value: ()).join(joint)
+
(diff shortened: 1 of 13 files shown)


mxcl:
Many thanks.

The `join` stuff is new, here's an example:


```swift

class Engine {
	static func make() -> Promise<Engine> {
		let (enginePromise, joint) = Promise<Engine>.joint()
		let cylinder: Cylinder
		
		firstly { () -> Void in
			// Engines are expensive to make, so we want
			// to initialize its components first.

			cylinder = Cylinder(explodeAction: {
			
				// the problem is we need access to the engine
				// every explosion
				
				// we *could* an IUO, but Swift is safe and there
				// are no guarantees about when this callback will
				// be called. Having an actual promise is safe.
			
				enginePromise.then { engine in
					engine.checkOilPressure()
				}
			})
		}.then {
			Ignition.default.start
		}.then { plugs in
			Engine(cylinders: [cylinder], sparkPlugs: plugs)
		}.join(joint)
		
		return enginePromise
	}
}
```

I hate this example, but actually I couldn't come up with a better one.

Perhaps that means the pattern is bad.

Anyway, this feature is “beta”, hence I added it without docs or announcement. I need to experiment with it, and I hope others might too, so we can figure out if it makes sense or not.

Certainly I have needed something like `join` many times.

zlangley:
Ah, I see, that makes sense. Thanks for the sample code! I think adding this as documentation will encourage people to experiment with the feature.

Will update.

zlangley:
Updating the docs now. Just so I'm clear, is this basically just a convenience for pending promises? As far as I can tell, you could have also written your example as:

```
class Engine {
    static func make() -> Promise<Engine> {
        let (enginePromise, fulfill, reject) = Promise<Engine>.pending()

        let cylinder = Cylinder(explodeAction: {
            enginePromise.then { engine in
                engine.checkOilPressure()
            }
        })

        firstly {
            Ignition.default.start()
        }.then { plugs in
            Engine(cylinders: [cylinder], sparkPlugs: plugs)
        }.then { engine in
            fulfill(engine)
        }.catch { error in
            reject(error)
        }

        return enginePromise
    }
}
```

mxcl:
> Just to be clear, this is basically just a convenience for specific use of pending promises?

Yes, true.

zlangley:
Cool, let me know if I should make any changes.

mxcl:
Terrific work, many thanks.

PR #578, opened 2016-10-30
Replace PMKJoint and associated methods with Promise.pendingWithPipe()

zlangley:
Since the use of this API is similar to that of Promise.pending(), it
makes sense to provide a similar API so that only one pattern needs to
be learned.

### DIFF
--- a/03_PendingTests.swift
+++ b/03_PendingTests.swift
@@ -2,49 +2,49 @@ import PromiseKit
 import XCTest
 
 class JointTests: XCTestCase {
-    func testPiping() {
-        let (promise, joint) = Promise<Int>.joint()
+    func testPipingFulfilledPromise() {
+        let (promise, pipe) = Promise<Int>.pendingWithPipe()
 
         XCTAssert(promise.isPending)
 
         let foo = Promise(value: 3)
-        foo.join(joint)
+        pipe(foo)
 
         XCTAssertEqual(3, promise.value)
     }
 
-    func testPipingPending() {
-        let (promise, joint) = Promise<Int>.joint()
+    func testPipingUnfulfilledPromise() {
+        let (promise, pipe) = Promise<Int>.pendingWithPipe()
 
         XCTAssert(promise.isPending)
 
         let (foo, fulfillFoo, _) = Promise<Int>.pending()
-        foo.join(joint)
+        pipe(foo)
 
         fulfillFoo(3)
 
         XCTAssertEqual(3, promise.value)
     }
 
-    func testCallback() {
+    func testPipingHandlerWithFulfilledPromise() {
         let ex = expectation(description: "")
 
-        let (promise, joint) = Promise<Void>.joint()
+        let (promise, pipe) = Promise<Void>.pendingWithPipe()
         promise.then { ex.fulfill() }
 
-        Promise(value: ()).join(joint)
+        pipe(Promise(value: ()))
 
         waitForExpectations(timeout: 1)
     }
 
-    func testCallbackPending() {
+    func testPipingHandlerWithUnFulfilledPromise() {
         let ex = expectation(description: "")
 
-        let (promise, joint) = Promise<Void>.joint()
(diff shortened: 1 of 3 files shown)


zlangley:
@mxcl Curious about your thoughts on this. `pendingWithPipe()` isn't a great name, but I didn't want to overload `pending()` and introduce ambiguities at the call site.

mxcl:
Yes, you’re right, this is the correct approach. The pattern is the same. Good spot.

Let's stew on the name a little.

zlangley:
A couple more thoughts on this.

We could overload `pending`. If you unpack the tuple, there's no ambiguity. It would, however, require you to specify the type if you didn't want to unpack it, and we'd still have to name a type alias for the (Promise, Pipe) variant.

```swift
let (promise, fulfill, reject) = Promise<Int>.pending() // fine
let (promise, pipe) = Promise<Int>.pending() // fine
let root = Promise<Int>.pending() // ambiguous, compile error
let root: PendingTuple = Promise<Int>.pending() // fine, but PendingTuple should be renamed since there are now two PendingTuples
```

There's also the option of having a `PendingPromise` type, which you can then choose how to resolve later.
```swift
let pendingPromise: PendingPromise<Int> = Promise<Int>.pending() // or would you just instantiate with PendingPromise<Int>() ?
pendingPromise.fulfill(3)
pendingPromise.reject(error)
pendingPromise.pipe(promise)
```

I think a previous version of PromiseKit may have had this `PendingPromise` type?

mxcl:
Yeah good idea, we could just drop a pipe into the pending. Personally I almost always unwrap the `pending` when assigning to it, which means I will need to learn to add an additional `_`, which is somewhat ugly… but has elegance.

I still prefer the previous route of having the pipe go into the chain rather than take a promise:

```
let promise = firstly {
    //…
}.then {
    //…
}

pipe(promise)

//versus:

firstly {
    //…
}.then {
    //…
}.pipe(promise)
```

For convenience (it's often painful to grab the promise chain instance), and because it reads better, “firstly, then, pipe”.

But open to discussion, ofc.

mxcl:
Another option:

```swift
let (promise, fulfill, reject) = Promise.pending()

firstly {
    //…
}.then {
    //…
}.join(fulfill, reject)
```

This is less efficient, but certainly a lot easier to understand.

zlangley:
Oh, I like that idea. One potential concern is that it's possible to misuse the API, and say, accidentally mix and match fulfill and reject blocks from different pending promises.

What do you think about the `PendingPromise` type approach I mentioned? I'm thinking `PendingPromise` is a subtype of `Promise` (maybe `Promise` becomes a protocol).

```swift
func make() -> Promise<Engine> {
    let enginePromise = PendingPromise<Engine>()

    firstly {
        // …
    }.then {
        // …
    }.pipe(enginePromise)

    return enginePromise
}
```

`PendingPromise` would also have `fulfill` and `reject` methods, and we'd deprecate `Promise.pending()`.

That is, we'd have something like:

```swift
class PendingPromise<T>: Promise<T> {
    func fulfill(value: T) { /* … */ }
    func reject(error: Error) { /* … */ }
}

extension Promise {
    func pipe(pendingPromise: PendingPromise<T>) { /* … */ }
}
```

zlangley:
Come to think of it, I guess we already have a pending promise type. There's no reason to unpack the tuple in your example, we could just do:

```swift
let promise = Promise.pending()

firstly {
    //…
}.then {
    //…
}.join(promise)
```

mxcl:
I haven't had any good ideas about this.

mxcl:
I explored having a single `pending` in the experimental branch. We should all check it out in use etc.

Not yet pushed this specific change, will soon.

Nothing is final, I'm working on the branch and then will try it out in an app.

mxcl:
The experimental branch drops joint and instead has a consistent `Sealant` for all methods of resolution.

zlangley:
Awesome!

zlangley:
Going to close this PR since it sounds like you found a better solution.

Issue #560, opened 2016-10-02
Consider providing overloads for tuple returns of promises


mxcl:

FilipZawada:
Can you post an example of desired functionality please?

mxcl:
```swift
asyncThing().then {
    let foo = promise1()
    let bar = promise2()
    return (foo, bar)
}.then { promise1Result, promise2Result in
    //…
}
```

FilipZawada:
Cool, we'll have to add type declaration I guess:

```
Promise(value: 0).then { _ -> (Promise<String>, Promise<String>) in
    let foo = Promise(value: "foo")
    let bar = Promise(value: "bar")
    return (foo, bar)
}.then { foo, bar -> Void in
    print(foo + bar) // foobar
}
```

I'll try to have PR ready by tomorrow.

mxcl:
Great, I think it'll make a good addition, provided it doesn't cause the Swift compiler to get confused by other usages. Which is my main concern. Swift already gets confused by PromiseKit enough.

FilipZawada:
As a starting base I made a very minimalistic [implementation](https://github.com/FilipZawada/PromiseKit/commit/3740e99f2aadb2062e2bdb97e92541dc335898c5).

And now I wonder about few things:
* I made a support only for 2 & 3-values tuples. I feel at least 5 values should be allowed, however as for now I wanted to be consistent with `when`.
* I can't think about any way of allowing infinite number of promises in a tuple (except hacking with dangerous unsafeBitCast).
* I haven't added tuples functionality to `firstly`, since `firstly` already doesn't match `then` functionality. I guess you kept it intentionally thin?
* I haven't touched objc at all. Should I provide an adequate implementation too?
* I've provided a very minimalistic set of tests. E.g. do you find a rejection test useful? (I can add it)

What are your thoughts?

FilipZawada:
P.S. I will add a missing test case when one of returned promise is a `self`

mxcl:
> I made a support only for 2 & 3-values tuples. I feel at least 5 values should be allowed, however as for now I wanted to be consistent with when.

By all means add support for more, and feel free to increase the size of when also.

> I can't think about any way of allowing infinite number of promises in a tuple (except hacking with dangerous unsafeBitCast).

I would be interested in seeing the code, but yeah, sounds binary unsafe.

> I haven't added tuples functionality to firstly, since firstly already doesn't match then functionality. I guess you kept it intentionally thin?

[...]

PR #593, opened 2016-11-21
#560 Promises tuple

FilipZawada:
Add support for promises returning tuples of promises (#560)

- Supported up to 5 promises in tuples
- Supported 4 & 5 promises in `when` -> `when(fulfilled: p1, p2, p3, p4, p5)`

I had to add new `pipe` function that executes on provided dispatch queue. That's because I re-use`when` internally, and it is always run on `zalgo`.

Sidenote:
I don't check if in returned tuple there is originating promise itself (A+ 2.3.1). That's would not be allowed by compiler:
```
var promise<(Int, Int)> = somePromise.then { _ -> (Promise<Int>, Promise<Int>) in
    return (promise, promise) // fails
}
```
unless you do:
```
var promise<(Int, Int)> = somePromise.then { _ -> (Promise<Int>, Promise<Int>) in
    return (promise.asVoid().then {0}, promise.asVoid().then {0})
}
```
which we don't prevent anyway.


FilipZawada:
I also added support for `firstly`

PR #597, opened 2016-11-29
add AnyPromise+Private.h import into PMKCallVariadicBlock.m

zayhero:
Header `AnyPromise+Private.h` is missing in `PMKCallVariadicBlock.m` which may cause compile error like this

```
Pods/PromiseKit/Sources/PMKCallVariadicBlock.m:114:16: error: implicit declaration of function 'PMKProcessUnhandledException' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        return PMKProcessUnhandledException(thrown);
```

This header already added into the master branch https://github.com/mxcl/PromiseKit/blob/master/Sources/PMKCallVariadicBlock.m#L4
But since we are using `swift-2.x` branch, so add the import into this branch as well.

@mxcl, could you take a look? Thanks

### DIFF
--- a/Sources/PMKCallVariadicBlock.m
+++ b/Sources/PMKCallVariadicBlock.m
@@ -2,6 +2,7 @@
 #import <Foundation/NSDictionary.h>
 #import <Foundation/NSError.h>
 #import <Foundation/NSException.h>
+#import "AnyPromise+Private.h"
 #import "NSMethodSignatureForBlock.m"
 #import <PromiseKit/Umbrella.h>
 #import <PromiseKit/PromiseKit.h>


mxcl:
What compiler throws this error? Since we didn't get this error when PMK3 was HEAD for 1.5 years.

zayhero:
@mxcl I am using BUCK to build this, which I think is using normal Xcode toolchain to build. 

Here is the LLVM version I got from `llvm-gcc`
```
$ llvm-gcc -v
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Is there a particular concern in adding this header?

mxcl:
I'll fix Travis first

zayhero:
@mxcl just want to check if you already fix travis? If yes, I could push the branch again to make the build green

mxcl:
Sorry not yet. I can restart the build once I have.

PR #601, opened 2016-12-02
Update Readme with new example

bitwit:
This is a copy paste of the example of how to create a new pending promise from `Promise.swift`

I was surprised that this sort of example isn't front and center in the readme or the website docs given that it is a common requirement and quite a familiar API coming from other languages like Javascript


mxcl:
Thanks and good point.

PR #602, opened 2016-12-03
Added Promise.void

### LINKED ISSUE #532: Get resolved instance of Promise<Void>
* PromiseKit version: 4.0.0
* Xcode version: Xcode8 GM
* Installation method:
  - [x] Carthage
  
I may be missing the obvious-- but how can I create a resolved instance of `Promise<Void>` ?

I know there is `public func asVoid()` but that assumes you have a Promise. What if I simply want a already resolved Promise without value?

`required public init(value: T)` doesn't work and `Promise()` doesn't appear to be working either.

Thanks for your help. Sorry for the possibly silly question!

ldiqual:
@mxcl Fixes #532, related to my comment here: https://github.com/mxcl/PromiseKit/issues/532#issuecomment-264600471

I understand now why you said you were getting a compiler error: I was using `.void` in a context where `T` is already inferred:

```
func doSomething() -> Promise<Void> {
    return .void
}
```

What I assume you did is this:

```
let promise = Promise.void
```

which does indeed cause a compiler error because `T` cannot be inferred. If we want to support this kind of declaration, maybe having a non-generic helper namespace like `PromiseUtils` would help since we could use `PromiseUtils.void`. It seems super overkill for a small syntactic sugar addition like this though.

Let me know if you'd like to go with the current change, or add it somewhere else, or discard this altogether. I'm not super attached to having it in `PromiseKit` since it's an easy extension to write, and on the other hand I feel like a lot of people might need this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /Cartfile.resolved
 /.build
 .DS_Store
+DerivedData/
--- a/Sources/Promise.swift
+++ b/Sources/Promise.swift
@@ -528,3 +528,19 @@ extension Promise where T: Collection {
         }
     }
 }
+
+extension Promise {
+    
+    /**
+     Returns a fulfilled Void promise.
+     
+     Usage example:
+ 
+     func doSomething() -> Promise<Void> {
+        return .void
+     }
+    */
+    public static var void: Promise<Void> {
+        return Promise<Void>(value: ())
+    }
+}
--- a/Tests/CorePromise/01_PromiseTests.swift
+++ b/Tests/CorePromise/01_PromiseTests.swift
@@ -75,6 +75,23 @@ class PromiseTests: XCTestCase {
 
         let bad = Promise(value: ()).then { Error.dummy }
     }
+    
+    func testVoidPromise() {
+        
+        let promise = Promise<Void>.void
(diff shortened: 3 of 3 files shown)


mxcl:
I'm not happy that it requires you to clarify the type to prevent an error, since the whole point is it has a specific type, it will just be confusing to users.

I'd be interested to see usage examples as to why people need this so much, in my projects I need pre-resolved promises only rarely.

ragboyjr:
@mxcl from my own projects, I use resolved promises a lot when i'm trying to short circuit a promise chain.

```swift
func doSomethingWithUser(user: User) -> Promise<Void> {
    if !user.canDoSomething() {
        return Promise<Void>();
    }
    // actually do something and return a promise that does work.
    return Promise<Void>() { resolve, reject in 
        // do something
    }
}
```

I don't know if this in correct usage of promises, but this is typically when I'd use a resolved promise.

mxcl:
@ragboyjr yeah this is my usage too.

For PMK 5 I have restored the short circuit, which has led to me having to make all other initializers more explicit or it can short circuit their closure inputs as the Promise’s type (!). However, it feels right.

mxcl:
Added in PromiseKit 4.3.0 thanks to changes in Swift 3.1

Issue #595, opened 2016-11-28
Adding @discardableResult to `then`


FilipZawada:
I'd expect promises to return @discardableResult, to avoid compiler warning.

Right now this:
```
promise.then {
  ...
}.then {
  ...
} 
```

Causes compiler to warn about unused result. To prevent you have to do:

```
_ = promise.then {
  ...
}.then {
  ...
} 
```

@mxcl What about me making a PR for this?

zlangley:
Dupes #549, #551, #561, #574, #584.

FilipZawada:
Ah sorry :-) At least it'd be work to re-work tests, so they don't spam with compiler warnings so much.

mxcl:
I think we need to some work here at least, eg. I use `after(interval: _)` a lot, there's no reason to warn for that. It cannot fail.

So maybe we should add an `UnfailablePromise` class.

Also, as an aside, I added a `fatalCatch()` chain terminator to my local branch, and use that quite a lot.

mxcl:
Also, during my use of PromiseKit, almost always I find that if I am getting this warning it means I am not handling errors. Only occasionally am I added a sub-chain where I know I am handling the error elsewhere and thus adding a `_ =` prefix is necessary and thus also a good indicator that I know the error is handled elsewhere.

Other than the above mentioned cases where the promise cannot fail.

Another thing on this, this warning almost makes the `UnhandledErrorHandler` unnecessary, since the Swift compiler is ensuring we handle errors. We may want to consider deprecating it, unless I'm missing something.

mxcl:
The other option of course is to add the `@discardableResult` annotation.

I feel this would lead to less good code for our users.

But it would lead to less tedious usage for our users.

Honestly, I feel how we have it *now* is more Swifty. However, we are not Apple, and our popularity to an extent depends on us not being tedious.

zlangley:
Since this is consistently brought up, we might want to add something about it to the FAQ. We could also mention your `fatalCatch()` suggestion. (Although I'm not sure people are checking the FAQ before creating issues here.)

I agree that something like `UnfailablePromise` could simplify a several APIs (`after`, non-throwing `recover`, non-throwing `UnfailablePromise.then`) and would therefore be worthwhile.

It would be nice if we could customize the unused result warning to indicate that you're expected to terminate promises with a catch, but I don't think that's currently possible.

FilipZawada:
[...]

Issue #584, opened 2016-11-08
@discardableResult and recover


gmckenzi:
* PromiseKit version: 4
* Xcode version: 8.1

I know that PromiseKit doesn't use discardableResult intentionally to encourage use of catch. Which is fine. But, why doesn't recover get the same treatment as catch? In other words, when I end a chain with recover, doesn't that satisfy the same design goal, and thus shouldn't recover be marked as discardableResult?

mxcl:
You don’t have to recover the error, you could still throw.

I guess we *could* have multiple versions `recover`, if you use the one that doesn't throw then the result could be discarded.

gmckenzi:
True enough. I hadn't considered that. It isn't how I'm using recover, but I understand your point. Carry on.

mxcl:
I use it that way quite regularly, only recovering from some errors.

Issue #574, opened 2016-10-28
Proposal to mark `then` with `@discardableResult`


johngibb:
* PromiseKit version: 4.0
* Xcode version: 8.0

I'm thinking the majority of functions, such as `then`, should be marked with `@discardableResult`, to avoid having to write code like:

```
_ = api.login(user: "test", password: "1234").then { user in
    ....
}
````

Happy to submit a PR if you agree!

mxcl:
Dupes #561, #551, #549, #548.

We consider this a feature, see other tickets for rationale. Thanks.

Issue #561, opened 2016-10-12
Issues with ending promise chains


AttilaTheFun:
* PromiseKit version: latest
* Xcode version: latest
* CocoaPods version: latest stable

I've struggled with this issue for a while, and it has become more annoying in swift 3. I'm not sure if i'm doing something wrong, but i can't cleanly terminate chains of promises. Take, for example, a simple request to fetch a string from the server and print it to the terminal. I'd like to be able to do the following:

```
firstly {
    return URLSession.shared.dataTask(with: request).asString()
}.then { string in
    print(string)
}
```

However, the compiler warns me about the unused result from `then`, and often fails to infer that the last `then` closure in the chain should return void. It insists I either return an AnyPromise, or decorate my closure params with the `(...) -> () in` syntax which is awkward. I would like to propose adding a version of then that does not return anything called "finally" or something to mark the end of a chain of promises. Its closure could always return void to help the type inferencer along.

nathanhosselton:
You're not handling errors. We think you should be, so we don't allow results to be discardable as our way of subtly pushing developers end all of their promise chains with a`catch`. Although you can also silence the warning by assigning the result to `_`. See also #548 #549. If you feel this is overbearing, discussion is welcome.

Also, we do have a "finally"! We call it `always`.

mxcl:
Closing, dupes #551, #549, #548.

As said, discussion welcome.

AttilaTheFun:
@nathanhosselton @mxcl I think the name `always` is confusing, as it's unclear whether it executes only when the chain of promises completes successfully, or if it fires regardless of whether the chain rejects or fulfills. I think firstly / finally as bookends of a promise chain make sense, and i believe in a previous promise kit release that was the case. I'm not sure what the motivation was behind renaming it. And I understand the motivation behind wanting people to handle errors, but often you really don't want to do anything with that error beyond maybe logging it. Maybe a compromise would be an alternative to `catch`, `drop` which just logs the error so you can cleanly and explicitly indicate that you don't care about the error. 
For example:

```
firstly {
    return URLSession.shared.dataTask(with: request).asString()
}.then { string in
    print(string)
}.drop()
```

mxcl:
It executes if the promise rejects and it executes when it fulfills.

I considered something like `drop`, but it seemed to me:

```swift
_ = firstly {
    return URLSession.shared.dataTask(with: request).asString()
}.then { string in
    print(string)
}
```

Was more “swifty”.

PR #605, opened 2016-12-14
Added @discardableResult to the then statements to silence the unused…

terwanerik:
… warnings (that do not dissapear with _ -> Void)

I'm not sure this is the preferred behaviour, but the 'result of .. unused' warnings where irritating me. Trying to silence 'm with `.then { _ -> Void in` wasn't working either.


mxcl:
Dupes #549, #551, #561, #574, #584 and #595.

PR #610, opened 2016-12-29
Fix promise name in Cache 'n' Fetch example code

interstateone:
I think this is change is correct but let me know if I'm instead misunderstanding how this _should_ work. 😅


mxcl:
Good catch & thanks!

PR #612, opened 2017-01-05
Add @discardableResult to catch

AnthonyOSX:
With PromiseKit 3 it is now possible to put always blocks at the end of the chain. The proposed change will remove the annoying  "Result to call is unused" warning from Xcode.


mxcl:
True, thanks.

PR #631, opened 2017-03-02
Adding void promise factory function.

drekka:
I thought I'd try adding a void pointer factory function as we discussed today in the chat sessions. As you say it cannot be added to the Promise class because of of the need to also specify the generic. However it appears to work quite well as a standalone function. ie.

    let vp = voidPromise()

The only reason to add this was that it seemed intuitive to use a function like over using an initialiser that takes a value. 

This request includes this code and a test in case you like it.

### DIFF
--- a/Sources/Promise.swift
+++ b/Sources/Promise.swift
@@ -2,7 +2,6 @@ import class Dispatch.DispatchQueue
 import class Foundation.NSError
 import func Foundation.NSLog
 
-
 /**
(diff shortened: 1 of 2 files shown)


drekka:
Cancelling . Just realised I've using different formatting.

mxcl:
Yes, it works as a standalone, but that doesn't seem intuitive to me and thus I debate its utility.

PR #642, opened 2017-03-26
add traditional chinese language support.

wangchauyan:
Hello, 

Just create a traditional Chinese README file, I know many people use this library, so, I think it would be nice to have traditional Chinese related document for developers who would like to quickly learn and use your library. 

Thanks a lot.


mxcl:
Much appreciated!

Issue #620, opened 2017-01-26
Wrong binary version number with Carthage


cikpis:
* PromiseKit version: 4.1.4
* Xcode version: Version 8.2.1 (8C1002)
* Carthage version: github "mxcl/PromiseKit" "4.1.4"

Hi guys, I'm having trouble to use your SDK with Usabilla one because looking into your binary through ```otool -L``` I got ```@rpath/PromiseKit.framework/PromiseKit (compatibility version 0.0.0, current version 0.0.0)```

That is wrong because the value should be at least ```1.0.0```

Any clue why?

I think the problem is because ```Compatibility Version``` and ```Current Library Version``` are not set and they should be set to the same value than the ```Build Version```. 

Looking forward to your answer.

mxcl:
I don’t know where these things are set in Xcode, do you?

cikpis:
In build setting if you look for ```Compatibility Version``` and ```Current Library Version``` you should find them.

cikpis:
After changing the ```Compatibility Version``` and ```Current Library Version``` manually to ```4.1.4```, the Usabilla SDK runs as expected. 

Please fix this issue which should be a critical one and it could be fixed in 1 sec ;)

Keep me updated, cheers!

mxcl:
K, will do that presently.

mxcl:
Pushed that fix and tagged 4.1.5. LMK if there's any further trouble.

cikpis:
Cool! Thanks!

PR #651, opened 2017-04-19
Set Compatibility version and Current version to 1

victorg1991:
I'm trying to distribute a binary which has this library as a dependency. I develop it using carthage

When I try to set up a project with my library using cocoapods and try to run it in the simulator it shows: Library version: XXX requires version 4.1.8 or later, but PromiseKit provides version 1.0.0

This is because cocoapods don't set the compatibility version and the current version (they reverted that change https://github.com/CocoaPods/CocoaPods/issues/3940 https://github.com/CocoaPods/CocoaPods/issues/3903

The reason of this change is because this seems to be the recommended approach 

You can see more related in this issue too (https://github.com/CocoaPods/CocoaPods/issues/3903)


mxcl:
We already changed this because of Carthage: https://github.com/mxcl/PromiseKit/issues/620

So, I worry about it.

victorg1991:
Oh sorry, I made a mistake, which I wanted to do is revert that change and use the default value which is that most of the frameworks use (instead of the current version of the framework)

mxcl:
I will aim to tag a release and pod today, but it may end up being tomorrow, pretty busy today.

victorg1991:
No problem at all, thank you so much! 💯

victorg1991:
Any updates? :)

mxcl:
Live.

victorg1991:
🎉

PR #656, opened 2017-04-25
Update NSURL in docs

zlangley:


mxcl:
Thing is though. It still is `NSURLEtc` for ObjC. And for Google juice. Dunno yet what *should* be chosen for non code samples.

zlangley:
I think all the changes made are either in Swift code or associated with Swift code.

README.markdown
```diff
@@ -189,7 +189,7 @@ github "PromiseKit/MapKit" ~> 1.0
 
 ## Choose Your Networking Library
 
-`NSURLSession` is typically inadequate; choose from [Alamofire] or [OMGHTTPURLRQ]:
+`URLSession` is typically inadequate; choose from [Alamofire] or [OMGHTTPURLRQ]:
```
mxcl: Well, this one isn’t associated with Swift code.

Sources/AnyPromise.h
```diff
@@ -41,13 +41,13 @@ typedef NS_ENUM(NSInteger, PMKCatchPolicy) {
 
  If you provide a block that takes a parameter, the value of the receiver will be passed as that parameter.
 
-    [NSURLSession GET:url].then(^(NSData *data){
+    [URLSession GET:url].then(^(NSData *data){
```
mxcl: This one is in an ObjC header…

README.markdown
zlangley: Since the sample code below is in Swift I don't find it strange to prefer `URLSession` here.

Sources/AnyPromise.h
zlangley: Yes, this should be reverted.

README.markdown
mxcl: `NSURLSession` is the class name IMO. When you talk about this class in person, what do you say? Everyone I know still says "NSURLSession" and I expect them to for quite some time to come.

zlangley: I would probably say `URLSession`, but I have no problem reverting this.

mxcl: Fair do’s. Like, I'm not sold on either way, so if you think it should be `URLSession` I'm cool with it.

PR #664, opened 2017-05-19
Build PromisesKit on Linux using Docker.

JaviLorbada:
As seen in #661 without CI on Linux, Promises Kit becomes a bit fragile on for its use on Swift Server Side project.

## What's in this Pull Request?

* The use of [docker-compose](https://docs.docker.com/compose/overview/) for easily build PromisesKit on a Linux container. For more information about Docker please see official [docs](https://www.docker.com/).

* Update Travis to not only build on mac OS, but also on Linux environment. So if one of the two fails the build fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,37 @@
-language: objective-c
-osx_image: xcode8
-
-env:
-  - PLATFORM=Mac
-  - PLATFORM=iOS NAME='iPhone SE'
-  - PLATFORM=tvOS NAME='Apple TV 1080p'
-  - PLATFORM=watchOS
-
-before_install:
-  - if [ -n "$NAME" ]; then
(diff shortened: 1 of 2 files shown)


JaviLorbada:
@mxcl I'm not sure if an update to the README with some instructions about docker-compose is required.

Also, I tried to keep it simple for now, so the in `docker-compose` and `travis.yml` files the swift image version it is specified, currently `swift:3.1`, and Travis uses Ubuntu 14.04. 

Let me know if we want to expand on that.

mxcl:
Really great!

> @mxcl I'm not sure if an update to the README with some instructions about docker-compose is required.

Probably not. The README is really for consumers of the library. Us maintainers can record the specifics of its maintenance here.

> Let me know if we want to expand on that.

Seems great and I'm very happy we can now ensure Linux will not break.

PR #682, opened 2017-07-09
Add DispatchTimeInterval to after() function

maguro:
While the use of TimeInterval is understandable from the perspective
of maintaining parity with the Objective-C function PMKAfter, it's not
quite as strongly typed as DispatchTimeInterval.  Also, what's really
going on is a dispatch after a deadline and so support for both seconds
and DispatchTimeInterval seem warranted.

### DIFF
--- a/Documentation/CommonPatterns.md
+++ b/Documentation/CommonPatterns.md
@@ -175,7 +175,7 @@ situations where tasks *must* be done sequentially; animation is a good example.
 
 ```swift
 let fetches: [Promise<T>] = makeFetches()
-let timeout = after(interval: 4)
+let timeout = after(seconds: 4)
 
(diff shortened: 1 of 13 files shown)


mxcl:
This is good.

However, I fear deprecating `after(interval:` but then adding a new function called `after(interval:` will just confuse Swift in use.

I'd like to do some testing in a playground first before we can decide.

maguro:
Maybe changing the deprecation comment to

    "Use after(seconds: TimeInterval) or after(interval: DispatchTimeInterval)"

Just a thought.  With that said, reverting the change back to `after(interval: TimeInterval)` wfm.  I just thought that it was more concise/consistent.

mxcl:
screen shot 2017-07-11 at 11 05 27 am

PR #684, opened 2017-07-11
ability to change default queue (server side swift)

k1313:


Sources/DispatchQueue+Promise.swift
```diff
@@ -38,16 +38,6 @@ extension DispatchQueue {
...
      */
-    class public final var `default`: DispatchQueue {
-        get {
-            return __PMKDefaultDispatchQueue()
-        }
-        set {
-            __PMKSetDefaultDispatchQueue(newValue)
-        }
-    }
+    static public var `default`: DispatchQueue = DispatchQueue.main
```
mxcl: This means the objc value for this function will be different to the Swift version.

k1313: Are these values the same in the current version?

mxcl: I don't understand this question, sorry.

mxcl: Yes they are the same, the code you removed is what makes them the same…

k1313: sorry, don't know obj-c :)

k1313: it's ok now?

mxcl:
PromiseKit 5 allows this.

PR #693, opened 2017-07-28
Fix syntax issues in `attempt` example

ryanschneider:
I just ended up using the `attempt` example in some code, but had to make some syntactical fix-ups to get it to work and figured I'd send them back your way:

- `after` takes a `TimeInterval`, not a `DispatchTimeInterval`
- The signature referred to `source` instead of `body` (which needed to be marked `@escaping`)
- Missing `{`

### DIFF
--- a/Documentation/CommonPatterns.md
+++ b/Documentation/CommonPatterns.md
@@ -241,17 +241,19 @@ need to cancel the underlying task!
 ## Retry / Polling
 
 ```swift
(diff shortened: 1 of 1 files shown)


mxcl:
Sorry cannot merge, the `after` is correct for the next version of PromiseKit which I intend to release quite soon. The rest is fine though.

mxcl:
https://github.com/mxcl/PromiseKit/blob/master/Sources/after.swift

ryanschneider:
Ok, updated to use the new `after(interval: DispatchTimeInterval)` signature.

mxcl:
You're a star ⭐️

PR #720, opened 2017-10-10
Fix potential error in sample code of Getting Started

juli1quere:
Hello, 

I’m discovering PromiseKit (it’s been on my todolist for a long time). So I’m reading the « Getting started », and I think there is a typo in the code to create our own Promise. Indeed, in the case of an error, the sample code call `fulfill(error)`. But I think it should be `reject(error)`. 

It may be the result of a misunderstanding from me. If it is: I’m sorry ! 

Thanks for your good work !


mxcl:
Good spot! Thanks.

PR #726, opened 2017-10-13
Compile for Android

rikner:
In order to compile PromiseKit for Android with [Swifty Robot](https://github.com/gonzalolarralde/swifty-robot-environment) just modify the Preprocessor Macro in Error.swift


mxcl:
4.4.3 pushed.

Issue #423, opened 2016-05-12
Resolving promises synchronously


juanjo-ramos:
I have a class that is a wrapper on top of an array in Swift where I want to guarantee the add, delete and fetch operations always happen on the main thread. So, I have something like this
```
private var array = [MyType]()

func addObject(myObject: MyType) -> Promise<Void> {
    return dispatch_promise(on: dispatch_get_main_queue(), body: { () -> () in
        firstly {
            self.fetchObject(myObject.id)
        }.then({ (existingObject) -> () in
            if let existingObject=existingObject {
                self.deleteObject(existingObject)
            }
            self.array.append(existingObject)
        })
    })
}

func deleteObject(myObject: MyType) -> Promise<Void> {
    return dispatch_promise(on: dispatch_get_main_queue(), body: { () -> () in
        if let i = self.array.indexOf(myObject) {
            self.array.removeAtIndex(i)
        }
    })
}
    
func fetchObject(id: String) -> Promise<MyType?> {
    return dispatch_promise(on: dispatch_get_main_queue(), body: { () -> MyType? in
        return self.array.filter({$0.id == id}).first
    })
}
```

My problem is on `addObject` that I would like the entire closure on `dispatch_promise` to resolve synchronously, blocking the main thread until they finish. So the promises in `firstly` and `then` inside `addObject` resolving synchronously and guarantee that no external call to `deleteObject` or `fetchObject` will be executed in between.

Is it possible to do that? Am I thinking about this completely wrong? I have tried exploring this with `when` but couldn't come up with any solution.

mxcl:
Blocking the main thread just means Apple will reject your app and it will crash (you get 20 ish seconds of block time before the kernel ejects you).

So…

mxcl:
We do provide `PMKHang()` in ObjC, but it is strictly meant for testing purposes.

juanjo-ramos:
I don't want to create a deadlock, just to resolve 2 promises synchronously and on the same thread. So, when calling `addObject` the method doesn't return until both the promises in the `firstly` and `then` clauses have been resolved.

mxcl:
So `when`? http://promisekit.org/when/

mxcl:
This is probably a solution for you:

```swift
var addObjectPromise = Promise()

func addObject() -> Promise<Void> {
    addObjectPromise = addObjectPromise.then {
        return fetchObject()
    }.then { existingObject -> Promise<Void> in        
        self.array.append(existingObject)
[...]

PR #741, opened 2017-11-30
Pure-swift hang() and tests.

abl:
Hi! I've been writing some commandline tools in Swift and wanted to leverage promises; I need to ensure the command doesn't terminate before the promises resolve. After reading #423, here's a ~port of PMKHang to Swift.

Tested using Xcode 8.3.3 / Swift 3.1 on macOS and Swift 4.0.2 on Ubuntu 14.04.

### DIFF
--- a/PromiseKit.xcodeproj/project.pbxproj
+++ b/PromiseKit.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0C42F31B1FCF86320051309C /* 03_HangTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0C42F3191FCF86240051309C /* 03_HangTests.swift */; };
+		0CC3AF2B1FCF84F7000E98C9 /* hang.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0CC3AF2A1FCF84F7000E98C9 /* hang.swift */; };
 		49A5584D1DC5185900E4D01B /* 03_WrapTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 49A5584B1DC5172F00E4D01B /* 03_WrapTests.swift */; };
 		6314112B1D5978D700E24B9E /* PromiseKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 63B0AC571D595E1B00FA21D9 /* PromiseKit.framework */; };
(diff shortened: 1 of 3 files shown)


mxcl:
PMK5 (recently released has `wait()`)

https://github.com/mxcl/PromiseKit/blob/master/Sources/Promise.swift#L73-L98

Now, it operates differently, and I don't know which approach is better.

Also, do you use PMK5 and not 4? I haven't pushed 5 to CocoaPods yet is why I ask. I planned to write a blog post and document it better first.

Anyway, this is much appreciated. What do you think? Do you want this in 4, maybe also in 5? Do you think the approach you have taken with runloops is the more sensible choice?

I'm not sure really, I added `wait()` to PMK5 as an experiment, so I don't have firm feelings on this matter.

abl:
> PMK5 (recently released has `wait()`)

Whoops, I forgot to mention that I tried `wait()` first and it deadlocked the main thread - for example:

```swift
func testWait() {
    let ex = expectation(description: "block executed")
    after(seconds: 0.02).done {
        ex.fulfill()
    }.wait()
    waitForExpectations(timeout: 0)
}
```

This makes sense; it's possible to remove the deadlock by using `done(on:, ...)` but if any part of your chain hits the main queue you're sunk...and the GCD API doesn't seem to offer a clean solution to this. (`dispatch_get_current_queue` is missed.)

> Also, do you use PMK5 and not 4?

Yes, I'm using PMK5 - figured it'd be best to start acclimating early. :)

> Do you want this in 4, maybe also in 5?

Haven't tested this revision against 4 - I had something similar that worked in 4 but was less elegant. Would be happy to have it in 5 and if you think it'd be valuable to backport to 4 I can revisit my original attempt.

> Do you think the approach you have taken with runloops is the more sensible choice?

I'm just copying your approach from PMKHang.m - I prefer your approach in `wait()`. `wait()` is conceptually cleaner with less overhead; `hang()` just happens to be useful when invoked from a non-runloop application.

mxcl:
K cool, then I think we need to offer both, so I'll figure out how to merge this.

mxcl:
Tests fail for iOS, tvOS and watchOS.

mxcl:
Will merge, when CI passes.

abl:
Sorry, I've been out sick.

Investigating iOS et al - practically, there's no reason to call `hang()` over `wait()` as all iOS/tvOS/watchOS apps will have runloops.

abl:
Ok, the issue is that Apple-y platforms (that is, not os(Linux)) use a different CFRunLoop setup. After this tweak the ATV and iOS tests pass locally - the watchOS tests won't build with/without my changes for some reason on Xcode 9.2.

Long term this may introduce other issues - other platforms being added to Swift (e.g. Haiku, Fuchsia) will have to be added to this conditional.

PR #759, opened 2018-01-11
Fixes Xcode 9.1 warnings

staminajim:
Hi!  Looks like XCode has tightened up their warnings again and you have to now specify that a function or block takes no args rather than just using `()`

Some background and links over here: https://stackoverflow.com/questions/44473146/this-function-declaration-is-not-a-prototype-warning-in-xcode-9


mxcl:
I pushed 1.7.4 to CocoaPods and tagged it for Carthage.

staminajim:
Awesome thanks @mxcl!

PR #790, opened 2018-02-21
Update readme.md to use `[weak self]`

scspijker:
We should use weak self, as creating a strong reference to a view controller (with it's view) is not a good practice for network requests that might take a while. User might have closed the view already.


mxcl:
Per our FAQ we don’t believe you need `weak self`:

https://github.com/mxcl/PromiseKit/blob/master/Documentation/FAQ.md#do-i-need-to-worry-about-retain-cycles

Thanks for the thought though.

PR #793, opened 2018-02-22
Update Installation.md

nathanhosselton:
⌘c+⌘v error


nathanhosselton:
keep forgetting to skip ci 😫😫😫

mxcl:
> I keep forgetting to skip ci 

We all do, which is why there are hundreds of tickets to Travis asking for an exclude list. They have done nothing about this for years. Which makes no sense since it’s their cost.

PR #796, opened 2018-02-25
Document 'finally'

ldiqual:
This documents the new `finally` method with a code example taken from http://promisekit.org/news/2018/02/PromiseKit-6.0-Released/


mxcl:
I'd rather have a smaller reference to finally as an aside in the existing ensure section. I don't feel it deserves a section to itself.

ldiqual:
@mxcl Done. Let me know if I should remove the example, I feel like it's helpful.

Issue #784, opened 2018-02-18
PromiseKit 6 Sources Lack Documentation


mxcl:
Much of it can be ported from PMK4.

PR #797, opened 2018-02-25
Partially document v6 source code

ldiqual:
Addresses some of https://github.com/mxcl/PromiseKit/issues/784

As advised, I took most of the documentation from the `v4` branch. I only added documentation for the core methods for now: `then / done / get / map / catch / recover / ensure` as well as the `Promise` class. Happy to helping documenting the rest in other PRs.

Let me know if this is on the right track.


mxcl:
You rule.

PR #804, opened 2018-03-01
Update attempt() for new syntax

sdgandhi:
Wasn't compiling because syntax changes


mxcl:
This makes the syntax for PromiseKit 4, our documentation represents PromiseKit 6. Thanks anyway, but this is the opposite of what we want.

PR #805, opened 2018-03-02
Add official Promises/A+ compliance test suite

### LINKED ISSUE #789: Have us run the Promises/A+ via a JavaScript bridge
We have re-implemented the suite as Swift, but this means:

1. We don't test Guarantee & AnyPromise (though since all the code is now in `Thenable` this probably is not a big deal).
2. Any updates to the suite do not come our way.

A JS bridge shouldn’t be a large undertaking and would give us the additional reassurance that we didn't make any mistakes.

ldiqual:
Fixes #789

This PR runs the official Promises/A+ compliance test suite in a JavaScriptCore context, against a JS-bridged `JSPromise` wrapper.

Basic info in the readme: https://github.com/ldiqual/PromiseKit/blob/199eee14d500e6368095534f246050c23171a701/Tests/JS-A%2B/README.md

### Promise bridge

This is the part I'm the least knowledgeable about: how to properly expose `Promise` to the JS context so that things like reference checking in 2.3.1 work. Here's how I've done it so far:

 - `JSPromise` class exposed to the JS context. It conforms to `JSExport`
 - `JSPromise` wraps a `Promise<JSValue>`
 - `JSPromise` has one method accessible from the JS context: `then(onFulfilled, onRejected) -> JSPromise`

The implementation of `then` is what I struggled with the most as I'm not too familiar with this way of passing two callbacks and getting a promise back. From my understanding of the spec, according to the state of the promise, `then` needs to call either `onFulfilled` or `onRejected` which can both return a value, a promise, or throw an error. So I pretty much did this:

```
class JSPromise: JSExport {

  let promise: Promise<JSValue>
  
  func then(onFulfilled, onRejected) -> JSPromise {
  
    let afterFulfill = promise.then(onFulfilled)
    let afterRejected = promise.then(onRejected)
    let newPromise = promise.tap { result in
      return result.isFulfilled ? afterFulfill : afterRejected
    }
    
    return JSPromise(promise: newPromise)
  }
}
```

This causes 2.3.1 to fail because `newPromise` is returned instead of `afterFulfill`. Not too sure how to work around that because I don't think we have a handler that allows looking at a promise result and return a promise, something like `then(result: (Result) -> Promise) -> Promise`.

Let me know if you have better ideas on how to expose the promise, or how to implement `then`.

### Test runs

Result: 260 passing, 2 failing. Takes about 10 seconds.

Example test run: https://gist.github.com/ldiqual/885da59e2bedc21ccc988b340c10c5ac

 - 2.3.3 is disabled
 - All 2.3.1 tests fail:

```
  1) 2.3.1: If `promise` and `x` refer to the same object, reject `promise` with a `TypeError' as the reason.
       via return from a fulfilled promise:
     Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  

  2) 2.3.1: If `promise` and `x` refer to the same object, reject `promise` with a `TypeError' as the reason.
       via return from a rejected promise:
     Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```


mxcl:
This is superb, you deserve massive kudos.

Is it possible to generate a JS `TypeError`? If so I think I can get 2.3.1 to pass.

mxcl:
BTW your implementation of the `then` that the A+ tests expect seems good to me. You’re using the right equivalents that JS will expect (ie. `then` and `recover) and there's no additional dispatches happening which may have made the behavior incorrect.

Well I guess there is one extra, the second `then` should probably be `on: nil` to prevent this.

We cannot test `catch` in the way A+ expects, so the only pity is not to test `map` but not much we can do really. I guess we could use `pipe` since it doesn't dispatch, then based on the input do the right method. But I think even if we decided this was more correct it can wait for this PR first.

So all that remains is fixing or disabling the tests and merging, really great work!

ldiqual:
@mxcl Awesome! Yeah I think we can generate a `new TypeError()` by using `JSUtils.sharedContext.evaluateScript()`. There might be less hacky ways but that's a start.

ldiqual:
@mxcl I added `JSUtils.typeError(message: String) -> JSValue`. Not sure I'll have time to get our `then` to throw it but it's there if you want to give it a try.

ldiqual:
@mxcl Got it to throw properly but I'm not super happy with the solution since I think it should be thrown from PromiseKit as opposed to `JSPromise` re-implementing the logic. If you can think of another way, let know or feel free to edit my branch.

All tests now pass. I wish there was more reliance on PromiseKit itself for some of the validation and return values because sometimes it feels like we're testing `JSPromise` itself while it's meant to be a super thin JS wrapper. But some of the tests are very much JS centric so I don't feel too bad about it.

Next week I'll look into while Travis is failing, I assume it's just because JSC is not linked against.

mxcl:
I figured that we could test for `PMKError.returnedSelf` and somehow get the JS typeError back to Javascript, but I'm just guessing about it. It seems like it would work since your JSPromise is just a thin wrapper around `Promise<T>` but maybe I'm missing something.

mxcl:
Remaining failures seem to be Swift 4 and Swift 3.1, I can probably fix these.

mxcl:
I don't have permission to push to your branch, but here's the diff for Swift 4:

```diff
diff --git a/Tests/JS-A+/MockNodeEnvironment.swift b/Tests/JS-A+/MockNodeEnvironment.swift
index 0cdab5f..60edf64 100644
--- a/Tests/JS-A+/MockNodeEnvironment.swift
+++ b/Tests/JS-A+/MockNodeEnvironment.swift
@@ -99,7 +99,12 @@ class MockNodeEnvironment {
             }
         }
         let timer = Timer.scheduledTimer(timeInterval: interval, target: block, selector: #selector(Operation.main), userInfo: nil, repeats: repeats)
-        let hash = UInt32.init(truncatingBitPattern: UUID().uuidString.hashValue)
+        let rawHash = UUID().uuidString.hashValue
+    #if swift(>=4.0)
+        let hash = UInt32(truncatingIfNeeded: rawHash)
+    #else
+        let hash = UInt32(truncatingBitPattern: rawHash)
+    #endif
         timers[hash] = timer
         return hash
     }
```

mxcl:
And here's the diff for Swift 3.1:

```diff
diff --git a/JSUtils.swift b/JSUtils.swift
index 16addfe..5c3afd7 100644
--- a/JSUtils.swift
+++ b/JSUtils.swift
@@ -102,3 +102,15 @@ enum JSUtils {
         print(lines)
     }
 }
+
+#if !swift(>=3.2)
+extension String {
+    func split(separator: Character) -> [String] {
+        return characters.split(separator: separator).map(String.init)
+    }
+
+    var first: Character? {
+        return characters.first
+    }
+}
+#endif
```

mxcl:
I made a PR: https://github.com/ldiqual/PromiseKit/pull/1

mxcl:
https://travis-ci.org/mxcl/PromiseKit/builds/348545693

mxcl:
Strangely it fails in the JS for iOS 8 and 9 (we only test these with Xcode 9.2 and Swift 4):

```
Test Case '-[PMKJSA_Tests.AllTests testAll]' started.
JS Exception at 82:5: Unexpected token '>'
	> eval@[native code]
	> ./index.js
	> __webpack_require__
	> global code
JS Exception at undefined:undefined: undefined is not an object
	> undefined
```

I can't see any reason for this off hand.

mxcl:
Can we increase the timeouts on the JS tests? Travis is stressed and we keep timing out for the final CI checkmark.

mxcl:
All pass! Anything you want before I merge? I was going to squash the merge unless you object.

ldiqual:
I'm a bit concerned that it will still timeout for PRs unrelated to this commit, preventing future contributors from getting their code merged quickly. Do tests now pass reliably or should I spend more time looking into it?

If it's all good, then 👍 Thanks a lot for reviewing!

Side question: are there other test suites that you're thinking about complying to? Something that'd test `map`, `when`, `Guarantee` etc.

mxcl:
> I'm a bit concerned that it will still timeout for PRs unrelated to this commit, preventing future contributors from getting their code merged quickly. 

It probably will at times, but all our tests fail occasional (Xcode sucks in CI), so it's not new, but if you know a way off hand to increase the timeout, that would be good. If not, I can look later, I'm sure some sed can do the trick!

> Side question: are there other test suites that you're thinking about complying to? Something that'd test map, when, Guarantee etc.

I don't know any other test suites. Do you?

PR #806, opened 2018-03-03
Make when(_ guarantees:) generic

samrayner:
I tried this out and it seems to work fine. Is there a reason you restricted `when()` to Void guarantees that I'm missing?

I'd like to also provide equivalents for the `when(_:_:_:...)` helpers but that doesn't seem possible without introducing ambiguity.


samrayner:
On second thoughts I suppose you'd want these in addition to the Void versions rather than instead of.

mxcl:
Yes, we should have this, mainly I haven't added it due to fear of ambiguity and the lack of an easy way to know in advance how it will affect the ambiguity of `when`

PR #823, opened 2018-04-04
The return of Zalgo

colinta:
Supports

```
conf.Q.map = nil
```

This makes all promises to be run immediately instead of using async.

### DIFF
--- a/Documentation/FAQ.md
+++ b/Documentation/FAQ.md
@@ -235,6 +235,17 @@ PromiseKit.conf.Q.map = .global()
 PromiseKit.conf.Q.return = .main  //NOTE this is the default
 ```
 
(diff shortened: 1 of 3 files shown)


mxcl:
Let’s do the `return` one too while we’re here. If possible.

colinta:
👍 

I just noticed `get` doesn't take a queue, is that deliberate?

```swift
    func get(_ body: @escaping (T) throws -> Void) -> Promise<T> {
        return map(on: PromiseKit.conf.Q.return) {
```

mxcl:
No this is also an omission.

colinta:
👍 and added documentation around setting those queues to `nil`.

PR #828, opened 2018-04-08
fix memory leak

kingslay:
Xcode 9.3
when in ObjC
`resolve([NSError errorWithDomain:@"" code:0 userInfo:nil]);`
the NSError will not call `dealloc`.
After change `obj is Error` to `obj is NSError || obj is Error`
the NSError will call `dealloc`.

### DIFF
--- a/Sources/AnyPromise.swift
+++ b/Sources/AnyPromise.swift
@@ -25,7 +25,7 @@ import Foundation
     @objc public func __thenOn(_ q: DispatchQueue, execute: @escaping (Any?) -> Any?) -> AnyPromise {
         return AnyPromise(__D: __AnyPromise(resolver: { resolve in
             self.__pipe { obj in
-                if !(obj is Error) {
+                if !(obj is NSError || obj is Error ) {
                     q.async {
                         resolve(execute(obj))
                     }
@@ -39,7 +39,7 @@ import Foundation
     @objc public func __catchOn(_ q: DispatchQueue, execute: @escaping (Any?) -> Any?) -> AnyPromise {
         return AnyPromise(__D: __AnyPromise(resolver: { resolve in
             self.__pipe { obj in
-                if obj is Error {
+                if obj is NSError || obj is Error {
                     q.async {
                         resolve(execute(obj))
                     }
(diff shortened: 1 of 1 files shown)


Sources/AnyPromise.swift
```diff
@@ -64,8 +64,8 @@ import Foundation
     /// Internal, do not use! Some behaviors undefined.
     @objc public func __pipe(_ to: @escaping (Any?) -> Void) {
         let to = { (obj: Any?) -> Void in
-            if let err = obj as? Error {
-                to(err as NSError)  // or we cannot determine if objects are errors in objc land
+             if obj is NSError || obj is Error {
+                to(obj as? NSError)  // or we cannot determine if objects are errors in objc land
```
mxcl: Use `as!`

Sources/AnyPromise.swift
```diff
@@ -39,7 +39,7 @@ import Foundation
     @objc public func __catchOn(_ q: DispatchQueue, execute: @escaping (Any?) -> Any?) -> AnyPromise {
         return AnyPromise(__D: __AnyPromise(resolver: { resolve in
             self.__pipe { obj in
-                if obj is Error {
+                if obj is NSError || obj is Error {
```
mxcl: This is unnecessary.

Sources/AnyPromise.swift
```diff
@@ -25,7 +25,7 @@ import Foundation
     @objc public func __thenOn(_ q: DispatchQueue, execute: @escaping (Any?) -> Any?) -> AnyPromise {
         return AnyPromise(__D: __AnyPromise(resolver: { resolve in
             self.__pipe { obj in
-                if !(obj is Error) {
+                if !(obj is NSError || obj is Error ) {
```
mxcl: This is unnecessary.

Sources/AnyPromise.swift
kingslay: I think that swift has special handling of Error that can cause problems with reference counting.

mxcl: `NSError` is defined as `: Error` so this particular check surely cannot make a difference. If it does then 100% we need a comment to justify this existence of this check.

mxcl:
Closing due to inactivity.

I require a test case that verifies this leak and proves the fix before I will apply this, preferably also an open bug against swift itself so we can track the fix there.

Issue #838, opened 2018-04-18
the closure for ensure() should return a Promise to make the containing chain composable


ilyathewhite:
PromiseKit 6 (but applies to 4 as well)

Even though the `ensure` operator returns a promise, it assumes that its closure is synchronous; calls it and then immediately returns the promise that it's being applied to. The assumption is that ensure is expected to run at the end of the chain, so it doesn't need to be asynchronous. This works if `ensure` is used only in that way but breaks if we want to turn the chain into a reusable (compound) promise.

I wrote more about it [here](https://blog.rocketinsights.com/promisekit-and-ensure/). It's possible to implement an overload of `ensure` this way using `recover`. I think that it would be nice to have this overload as part of PromiseKit.

```
extension Promise {
    /// Boolean state that is used for sharing data between
    /// promises. It needs to be a class because a struct would
    /// be just copied.
    private class BoolBox {
        /// The stored value
        var value: Bool

        init(_ value: Bool) {
            self.value = value
        }
    }

    /**
     The provided closure executes when this promise resolves.

     This variant of `ensure` executes just as any other `then`
     clause which allows the provided closure to have
     ascynchronous code. Unlike `then`, the closure executes on
     either fullill or reject (same as `ensure`). If the closure
     is rejected, this rejects the containing chain as well.

     - Parameter on: The queue to which the provided closure dispatches.
     - Parameter execute: The closure that executes when this promise fulfills.
     - Returns: A new promise that resolves when all promises
     returned from the provided closure resolve.
     */
    public func ensureAsync(on q: DispatchQueue? = conf.Q.return, execute body: @escaping () -> Promise<Void>) -> Promise {
        let executedBody = BoolBox(false)

        return self.then(on: q) { value -> Promise in
            executedBody.value = true
            return body().then(on: q) { () -> Promise in
                return .value(value)
            }
        }
        .recover(on: q) { (error) -> Promise in
            if !executedBody.value {
                return body().then(on: q) { Promise(error: error)}
            }
            else {
                return Promise(error: error)
            }
        }
    }
}
```

mxcl:
Yes I've needed something like this before, seems like this would be enough:

```swift
func ensure(on: DispatchQueue? = conf.Q.return, _ body: () -> Promise<Void>) -> Promise {
    let promise = Promise<T>(.pending)
    pipe { result in
        on.async {
            body().pipe {
                promise.seal(result)
            }
        }
    }
    return promise
}
```

[...]

PR #841, opened 2018-04-20
Add ensureThen

mxcl:
Couldn’t have two variants of `ensure` without the dreaded ambiguity issues.

Sadly Swift cannot infer the type for multiline closures where there are two possible functions for it to match. :(

Refs #838

### DIFF
--- a/Sources/Catchable.swift
+++ b/Sources/Catchable.swift
@@ -150,6 +150,38 @@ public extension CatchMixin {
     }
 
     /**
+     The provided closure executes when this promise resolves, whether it rejects or not.
+     The chain waits on the returned `Guarantee`.
+
+         firstly {
+             setup()
+         }.done {
+             //…
+         }.ensure {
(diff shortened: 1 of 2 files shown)


Sources/Catchable.swift
```diff
@@ -149,6 +149,38 @@ public extension CatchMixin {
...
 
+    /**
+     The provided closure executes when this promise resolves, whether it rejects or not.
+     The chain waits on the returned `Guarantee`.
+
+         firstly {
+             setup()
+         }.done {
+             //…
+         }.ensure {
```
ilyathewhite: this is a typo -- should be `ensureThen` after the rename

ilyathewhite:
Thank you! Also, very impressive how quickly this was added.

Sources/Catchable.swift
mxcl: Oops thanks.

PR #856, opened 2018-05-15
Fix build issue in hang.swift

ephemer:
Android does not have `CFRunLoopMode.xyz` (which breaks builds for that platform). Android is also a Linux variant, so this change fixes the semantic / functional issue there too.

It's possible this should be a defensive check on `#if os(iOS) || os(macOS) || os(tvOS)` instead, because the same issues apply to all non-darwin platforms. But personally I'd be fine with this as is.

### DIFF
--- a/Sources/hang.swift
+++ b/Sources/hang.swift
@@ -13,7 +13,7 @@ import CoreFoundation
  - See: `wait()`
 */
 public func hang<T>(_ promise: Promise<T>) throws -> T {
-#if os(Linux)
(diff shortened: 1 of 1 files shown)


ephemer:
Hi @mxcl we are working on a solution for a CI for Android, it's pretty specific to our use case for now but could be interesting to generalise it in the future. I'll keep you in mind for future updates

mxcl:
That would be amazing! Thanks!

Issue #825, opened 2018-04-06
Build error in AnyPromise.m when installing PromiseKit 6 as a static library via CocoaPods


nolanw:
Hello! I'm using PromiseKit 6.2.4, CocoaPods 1.5.0, and Xcode 9.3.

CocoaPods 1.5.0 introduced support for Swift static libraries in the Podfile. That is, the CocoaPods end-user can decide whether to configure Swift dependencies as dynamic or static libraries.

When configured this way, PromiseKit 6.2.4 fails to build. Line 1 of `AnyPromise.m` reads:
```objc
#import <PromiseKit/PromiseKit-Swift.h>
```
and the compile error I get is
> /Users/nolanw/Desktop/Updog/Pods/PromiseKit/Sources/AnyPromise.m:1:9: 'PromiseKit/PromiseKit-Swift.h' file not found

if I change that line of `AnyPromise.m` to instead read
```objc
#import "PromiseKit-Swift.h"
```
then the build succeeds. Here is my `Podfile` in all its glory:

```ruby
target 'Updog' do
  pod 'PromiseKit'
end
```

I realize that using a system include is the recommended approach in Apple's docs for using Swift code in Objective-C in a Framework target, and I don't know what the ramifications would be for unilaterally changing to a user include. (I'm guessing the build would then fail for other installation methods.)

It's entirely possible that CocoaPods should be doing something with the build settings (or maybe we could add something to PromiseKit's `.podspec` to fiddle with build settings) so that the system include succeeds with this configuration? I don't mind opening an issue at CocoaPods if we think it's on them.

nolanw:
This seems to work ok as a replacement for line 1 of `AnyPromise.m`:
```objc
#if __has_include("PromiseKit-Swift.h")
    #import "PromiseKit-Swift.h"
#else
    #import <PromiseKit/PromiseKit-Swift.h>
#endif
```
or at least it still builds ok when installed via CocoaPods both as a dynamic library and a static library.

Kinda ugly but honestly I'm a bit shocked it's the only change needed to get a static library building under CocoaPods. Happy to open a PR if this looks workable!

nolanw:
OK, I can confirm that
```objc
#import "PromiseKit-Swift.h"
```
fails when installed via Carthage. I can confirm that the `__has_include` approach succeeds when installed via Carthage and when installed as dynamic and as static via CocoaPods.

Sorry for the rapid fire replies. I'm done for a little while :)

nolanw:
[...]

PR #859, opened 2018-05-16
Fixed build error in AnyPromise.m

evgenich95:
When installing PromiseKit 6 as a static library via CocoaPods.
The problem is described in the following issue - https://github.com/mxcl/PromiseKit/issues/825.


mxcl:
We have decided not to do this, pending advisement from CocoaPods if we should do anything or not.

Like I said in #825, your fix…

> Seems safe, but I am hesitant, I have applied similar hacks before and then had three new bugs show up after release.

I do not want to break a whole bunch of other stuff.

mxcl:
Reopening pending my question in the CocoaPods thread.

mxcl:
Yes, will try!

mxcl:
done

Issue #833, opened 2018-04-12
[objc] Generic type annotation for PMKPromise


theodelrieu:
Hello,

I'm using PromiseKit v1.7.6, and I have `PMKPromise*` everywhere in my APIs.

Users must rely on the documentation to know what is the type of the promise.

I would really like to have:

```objc
- (nonnull PMKPromise<NSString*>*)foo;
```

I don't know if this is implementable for types that are not `NSArray/NSString` etc...

But if it's possible that would bring a lot of clarity in APIs. Also, adding that to `AnyPromise` might help Swift users too.

What do you think?

mxcl:
I expect it can be done, but I do not have experience with it. PR welcome.

theodelrieu:
I won't be available for two weeks, but I'll try to do that after.

mxcl:
Closing since this is a feature-request, the PR is welcome but I do not leave feature requests open, the tracker is for bugs.

PR #862, opened 2018-05-24
add lightweight generics support

theodelrieu:
Hello,

This is a follow-up of #833, adding the ability to indicate `PMKPromise` result type.
This has no impact on runtime whatsoever, I followed [this blog post](https://miqu.me/blog/2015/06/09/adopting-objectivec-generics/) to implement it.

As you can see it is as straightforward as it can be, although there might be some way to improve this and use annotations inside the library's code.
I have not read the whole code base, so you're the one that knows @mxcl :)

From what I've understood, you're using `id` a lot, so determining the result type of a promise chain seems unfeasible.

### DIFF
--- a/Sources/PromiseKit/Promise.h
+++ b/Sources/PromiseKit/Promise.h
@@ -18,7 +18,7 @@
  @see [PromiseKit `then` Guide](http://promisekit.org/then/)
  @see [PromiseKit Chaining Guide](http://promisekit.org/chaining/)
 */
-@interface PMKPromise : NSObject
+@interface PMKPromise<ResultType> : NSObject
 
 /**
  The provided block is executed when its receiver is resolved.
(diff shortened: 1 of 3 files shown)


mxcl:
Is this a breaking change for consumers? Seems so since you had to change the test sources.

theodelrieu:
No, I did change the tests to show that it works, but this can be reverted. Objc generics are almost only cosmetic. The compiler can emit warnings though:

```objc
PMKPromise<NSNumber*>* p;
PMKPromise<NSString*>* p2 = p; // warning emitted
```

The warnings can be suppressed by using `__covariant`, as explained in the top comment of [this article](http://drekka.ghost.io/objective-c-generics/). But I believe they are quite valuable.

I don't know if promises using `PMKManifold` can be annotated, that's a limitation of this PR.
I would need to look further, but this looks like a good start.

mxcl:
K, thanks for clearing that up!

mxcl:
> From what I've understood, you're using id a lot, so determining the result type of a promise chain seems unfeasible.

The return type for all the handler blocks is inferred, but maybe something can be done in a later PR, not sure.

PR #871, opened 2018-06-09
Guarantee thenMap -> Guarantee

drekka:
Proposed addition to Guarantee to provide a `thenMap` function that returns a Guarantee instead of a promise.


mxcl:
The test needs to use `XCTestExpectation`s or it may not finish.

PR #872, opened 2018-06-09
Guarantee.thenMap which returns a Guarantee

drekka:
This PR adds a `thenMap` function to Guarantee that returns a Guarantee instead of a Promise.

Hopefully I've got it right. I've also added a test which passes.


drekka:
Updated the test to wait for the results.

PR #876, opened 2018-06-12
Accept a queue in finally()

mayurdzk:
@mxcl I'm not entirely sure of the test I've included in the test suite. Since I saw two tests in the `testFinally()` test, I added the test for checking the DispatchQueue there too.

Let me know if there's a better way you'd like me to structure this.

### DIFF
--- a/Sources/Catchable.swift
+++ b/Sources/Catchable.swift
@@ -45,8 +45,10 @@ public class PMKFinalizer {
     let pending = Guarantee<Void>.pending()
 
     /// `finally` is the same as `ensure`, but it is not chainable
-    public func finally(_ body: @escaping () -> Void) {
-        pending.guarantee.done(body)
+    public func finally(on: DispatchQueue? = conf.Q.return, flags: DispatchWorkItemFlags? = nil, _ body: @escaping () -> Void) {
+        pending.guarantee.done(on: on, flags: flags) {
+            body()
+        }
     }
 }
 
--- a/Tests/CorePromise/CatchableTests.swift
+++ b/Tests/CorePromise/CatchableTests.swift
@@ -1,18 +1,23 @@
 import PromiseKit
+import Dispatch
 import XCTest
 
 class CatchableTests: XCTestCase {
 
     func testFinally() {
+        let finallyQueue = DispatchQueue(label: "\(#file):\(#line)", attributes: .concurrent)
 
-        func helper(error: Error) {
+        func helper(error: Error, on queue: DispatchQueue? = nil, flags: DispatchWorkItemFlags? = nil) {
             let ex = (expectation(description: ""), expectation(description: ""))
             var x = 0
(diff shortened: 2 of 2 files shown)


mxcl:
This is a commonly requested feature that we have decided to reject. Sorry. Justification can be found in the ticket tracker.

mxcl:
Oh my bad, I thought this was for `firstly`.

Sources/Catchable.swift
```diff
@@ -45,8 +45,11 @@ public class PMKFinalizer {
     let pending = Guarantee<Void>.pending()
 
     /// `finally` is the same as `ensure`, but it is not chainable
-    public func finally(_ body: @escaping () -> Void) {
-        pending.guarantee.done(body)
+    public func finally(on: DispatchQueue? = conf.Q.return, flags: DispatchWorkItemFlags? = nil, _ body: @escaping () -> Void) {
+        let fp = pending
+        on.async(flags: flags) {
```
mxcl: Unfortunately this doesn’t work as expected. It immediately dispatches to this queue, we need to pass the queue and flags to the `done`.

Also the test didn't flag this, so we should make it so it does. The reason it doesn't is your check doesn't actually verify which queue we are on, use `dispatchPrecondition(.onQueue(q))` to detect this.

mayurdzk:
@mxcl I've addressed the feedback.

I don't have a way to check whether the CI tests pass/fail on Linux other than to wait for the results.

Also, let me know if we need to address the problem you mentioned... 
> Also the test didn't flag this, so we should make it so it does.

...apart from just using the `dispatchPrecondition(.onQueue(q))` check. (I'm not sure using just this precondition would ensure that a closure passed in the `finally` block truly is executed after every other closure in the promise. I'm open to adding more tests to validate this condition. Let me know your thoughts on this.)

mxcl:
Like, I said, the fix for Linux is `import Dispatch` at the top.

>  (I'm not sure using just this precondition would ensure that a closure passed in the finally block truly is executed after every other closure in the promise chain. I'm open to adding more tests to validate this condition. Let me know your thoughts on this.)

We have other tests that validate this so I don't consider it necessary, OTOH we have other tests that do this so you can just copy and paste.

Sources/Catchable.swift
mxcl: > Thanks for the feedback 😃. I tried using dispatchPrecondition(condition: .onQueueAsBarrier(q)) since it was used elsewhere in tests but it wouldn't pass

Well, this proves that it is not executing on the right queue. The assert passes because it is essentially testing if `1 == 1`.

mayurdzk:
> the fix for Linux is `import Dispatch` at the top.

Oops, I remember adding it but it must've got undone. Sorry about that. 

> we have other tests that do this so you can just copy and paste.

Will do!

PR #881, opened 2018-06-17
Documentation copy editing, add more RxSwift info, fix asVoid() example

GarthSnyder:
This is mostly just editing for grammar, consistency, and flow. A couple of points will need review, though:

1) I added a couple more sentences about Swift type inference to the Troubleshooting section. This is just general Swift info, nothing really specific to PromiseKit.

2) I edited one of the code examples about `asVoid()` in the Appendix. The original code was:

```
let p1 = foo()
let p2 = bar()
let p3 = baz()
//…
let p10 = fluff()

when(fulfilled: p1, p2, p3, /*…*/, p10).then {
    let value1 = foo().value!  // safe bang since all the promises fulfilled
    // …
    let value10 = fluff().value!
}.catch {
    //…
}
```

It seemed suspicious because a) this is described as an example that shows `asVoid()` being essential, but it does not call `asVoid()`, and b) if `foo()` etc produce promises, then doesn't calling those functions again just produce new, independent promises?  This was my guess at the intended code:

```
let p1 = foo()
let p2 = bar()
let p3 = baz()
//…
let p10 = fluff()

when(fulfilled: p1.asVoid(), p2.asVoid(), /*…*/, p10.asVoid()).then {
    let value1 = p1.value!  // safe bang since all the promises fulfilled
    // …
    let value10 = p10.value!
}.catch {
    //…
}
```

3) I added an in-line answer to the FAQ question about RxSwift. Some of the text draws from the existing trouble ticket, but most is new, although I think it's consistent with the material that is already around. This answer may be too long given the generally brief style of the documentation. Feel free to trim, or I'll do it if you prefer. I do think this is a common query for people new to both libraries.

### DIFF
--- a/Documentation/Appendix.md
+++ b/Documentation/Appendix.md
@@ -1,6 +1,6 @@
-# Common Misusage
+# Common misusage
 
-## Doubling Up Promises
+## Doubling up promises
 
 Don’t do this:
 
@@ -37,9 +37,9 @@ func toggleNetworkSpinnerWithPromise<T>(funcToCall: () -> Promise<T>) -> Promise
 You already *had* a promise, you don’t need to wrap it in another promise.
 
 
-## Optionals in Promises
+## Optionals in promises
 
-Mostly when we see `Promise<Item?>` it implies a misuse of promises, for
+When we see `Promise<Item?>`, it usually implies a misuse of promises. For
 example:
 
 ```swift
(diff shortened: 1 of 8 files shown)


mxcl:
Thanks, but why did you change the capitalization of all the headers to non-title case?

GarthSnyder:
It's fine either way and style guides vary on what they recommend, but it should be consistent throughout. 

I standardized them to sentence case largely on the general principle that extra formatting that doesn't aid readability can be removed. Markdown renders the headings pretty unmistakably, so there's no need for additional "specialness" to call attention to the fact that they're headings. In addition, many of the titles mention capitalized class names, and sentence case helps those stand out more (e.g., is the heading talking about a Promise-class object specifically or just a generic English "promise"?). 

Obviously, these are minor considerations. I'm happy to resubmit with any convention you prefer if you feel strongly about it.

There are a few other things in this category, too. I added Oxford commas, for example. There's nothing wrong with not having them; I just wanted to be sure they were consistent and that's the convention I'm more used to.

mxcl:
You are not going to be here next week, or next month or next year.

Let’s keep things my style. Thanks.

mxcl:
Thank you for this thorough work and thoughtful work.

PR #906, opened 2018-07-30
Add async as a Promise class func

ArifADS:
Just a quicker way to make an async function without having to directly use `DispatchQueue.async(PMKNamespacer)`.

The fun part of this extension is that you can use `.async { }` in the return of a function just like with `.value()`, making the code more readable. 

Example: 
```swift
func qResponse<T>() -> Promise<QResponse<T>> {
        return .async {
            let (data, res) = try self.response().wait()
            var qRes = try JSONDecoder().decode(QResponse<T>.self, from: data)
            qRes.status = (res as? HTTPURLResponse)?.statusCode ?? -1
            return qRes
        }
}
```

Happy to learn if there is a cleaner way I'm not seeing.

### DIFF
--- a/Sources/Promise.swift
+++ b/Sources/Promise.swift
@@ -102,6 +102,23 @@ public class Promise<T>: Thenable, CatchMixin {
 }
 
 public extension Promise {
+    
+    /**
+     Shorter way to make an async function
+
+     func myAsynFunc() -> Promise<Int> {
+         return .async {
+             let myNumber1 = try self.myPromisedCalulation().wait()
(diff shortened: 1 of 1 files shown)


ArifADS:
I think there should be a way to use the `.wait()` in a faster, more readable way than inside a `DispatchQueue.global(qos: qos.qosClass).async(.promise) { }` and conserving intuitiveness.

What do you think?

Thank you for this project.

mxcl:
I would like something, but I've not come up with anything so far.

PR #913, opened 2018-08-19
Test 3.2 on Linux (we weren't actually)

mxcl:


mxcl:
Closing because travis didn't run

PR #915, opened 2018-08-24
Fix spacing in the example code in README

mkj-is:
I noticed there is missing space in the second example line in README.md. If it was anywhere else (not on the front page) I would not be making such a small and silly pull request.


mxcl:
Thanks but this is deliberate, it's Ruby style inline closures that I prefer.

PR #923, opened 2018-09-19
Update Package.swift

heyzooi:
Updating Package.swift to add support for Swift 4.2

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.0
+// swift-tools-version:4.2
 
 import PackageDescription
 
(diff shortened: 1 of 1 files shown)


mxcl:
This breaks our support for some Swift versions

heyzooi:
@mxcl it seems possible... https://github.com/apple/swift-protobuf/blob/master/Package%40swift-4.2.swift

I will try that soon and push new commits to this PR

mxcl:
Would be good if we can add a travis test to verify we are actually getting 4.2 out of the swift compiler. If you can't think of a way, I'll merge and I can try to figure it out at a later time.

heyzooi:
@mxcl I'm not sure how this could be done. One way maybe would be replace the `swiftLanguageVersions` to the old format `[3, 4]` and check if it fails. Do you have any suggestion how to do it?

mxcl:
I published 6.4.1

Issue #928, opened 2018-09-25
Support for swift 4.2 


techmehra:
[PLEASE READ THE TROUBLESHOOTING GUIDE](https://github.com/mxcl/PromiseKit/blob/master/Documentation/Troubleshooting.md).

---

You read the guide but it didn’t help? OK, we’re here to help.

* Please specify the PromiseKit major version you are using
* [Please format your code in triple backticks and ensure readable indentation](https://help.github.com/articles/creating-and-highlighting-code-blocks/)

mxcl:
We already support Swift 4.2. Of you have a specific issue please open a new ticket.

ghowen:
@mxcl I just updated a project to the latest pods, and Xcode 10 still claims that there is a conversion to Swift 4.2 available for the PromiseKit.framework. It's just a warning that Swift code in the framework has not been converted to Swift 4.2 yet, but the library runs without a problem. Maybe this is what @techmehra meant.

mxcl:
I ported everything months ago, but maybe I didn’t update the UIKit extension.

You probably don’t even use it specify your pod request without as shown in the README.

mxcl:
K, I did indeed port this: 

https://github.com/PromiseKit/UIKit/commit/9c686799d23b12d01596ec1a0844265511acb586

And it's part of PromiseKit 6.4.0, so check you are up-to-date, if you still have problems then we will have to investigate further, but very much looks like issue with your cocoapods checkout to me.

ghowen:
That is indeed weird. I am on PromiseKit 6.4.0 and if I open the pod and look at the source I can see the `#if swift(>=4.2)` directive. But Xcode 10 GM still wants to convert the part after the `#else` to Swift 4.2 which does not make any sense really. TBH I'm a bit at a loss now.

mxcl:
Most likely because Cocoapods generates the `Pod.xcodeproj` for Swift 4.0 and Xcode is not being very smart here.

I imagine there's a ticket open to get pod to generate its project for 4.2 that you can subscribe to.

techmehra:
![screenshot 2018-09-27 at 10 37 31](https://user-images.githubusercontent.com/18284379/46124374-5d5f7580-c241-11e8-968a-f0c9ca10a893.png)

mxcl:
As already stated this is Cocoapods’ fault, I suggest opening a ticket there. PromiseKit is Swift 4.2 ready and has been for months.

Kaspik:
[...]

PR #936, opened 2018-10-08
Update PromiseKit.podspec

pmusolino:
Fixed Xcode warning relative to this https://github.com/mxcl/PromiseKit/issues/928#issuecomment-424960602

### DIFF
--- a/.github/PromiseKit.podspec
+++ b/.github/PromiseKit.podspec
@@ -23,7 +23,7 @@ Pod::Spec.new do |s|
   
   # CocoaPods documentation (like usual) is unclear about the
   # consequences of setting this; we support 3.2 THROUGH 4.2
-  # s.swift_version = '4.0'
+  s.swift_version = '4.2'
 
   # CocoaPods requires us to specify the root deployment targets
   # even though for us it is nonsense. Our root spec has no


mxcl:
I deliberately removed this. Cocoapods is broken. They have tickets open for a true fix where I can specify the range of Swift versions I support. If I don’t remove this we don’t sctuslly support all versions with Cocoapods and are lying.

mxcl:
That thread has a fix you can apply in your pod file. But even better stop using Cocoapods it’s a pos. Or open a ticket and tell them to fix their bullshit.

mxcl:
Like there is no reason whatsoever that Cocoapods doesn’t generate a project with this flag cleared so xcode doesn’t prompt you to upgrade the dependencies swift version. They are just lazy developers who have made a terrible piece of software.

mxcl:
Notably Alamofire also does not set this for the same reasons. Setting it to 4 was mostly acceptable IMO, but setting it to 4.2 to fix this warning would limit PMK a lot, so I just removed it instead, since the warning would be there for everyone on Xcode 10 anyway.

I updated our documentation to provide the “fix” and am watching the ticket to add version ranges to CocoaPods and may even step up and do the damn work.

pmusolino:
Thanks @mxcl for clarifying the situation.

mxcl:
Update: I decided to try and add the feature myself, so followed the “getting started for contributors” guide for CocoaPods, the provide a script to setup the environment, it took 40 minutes to run (wtf?) and failed. So fuck that shit. CocoaPods is awful.

PR #946, opened 2018-10-22
Fixed mistake with barrier on serial queue

djmixroman:


mxcl:
Seems like we should make the q concurrent instead, the only other use is a “read” operation.

djmixroman:
> Seems like we should make the queue concurrent instead, the only other use is a “read” operation.

Are u want to use concurrent queue.sync with barrier?

mxcl:
The queue should be iniatlized concurrent.

mxcl:
QoS utility seems wrong, this is often a user-initiated situation, where scrolling will trigger the need for an image to download. I would just leave it as `default` since there are many various situations where the QoS could be different, even where eg. you are pre-loading the images so you go with `utility` but then the user scrolls that cell into view and now you need it with `userInitiated` but this caching system has no way to upgrade the existing fetch to be a higher QoS.

PR #955, opened 2018-11-05
Fix typo

rehatkathuria:
I'm assuming this is a typo.

### DIFF
--- a/Documentation/CommonPatterns.md
+++ b/Documentation/CommonPatterns.md
@@ -198,7 +198,7 @@ foo.done {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rehatkathuria:
First of all, thank you to everyone for their hard work on PromiseKit. It's become an invaluable asset to asynchronous architecture that I write. Secondly, this pull-request might be moot but my understanding between `then` and `when` made me think this was a typo. Please correct me if I'm wrong!

mxcl:
No, it is correct as is. Usually you want `when` and not to use sequential thens.

Issue #639, opened 2017-03-20
Allow "filtered catch" to handle specific error types


ldiqual:
As described in the following Gitter conversation: https://gitter.im/mxcl/PromiseKit?at=58cf310a2215a08f04a3f9c2

I would like to have a `catch` per error type that I expect my promise to throw. Something similar to the way Bluebird (a popular JS Promise library) does it: http://bluebirdjs.com/docs/api/catch.html#filtered-catch

```
getAccount()
.catch(AccountNotFoundError) { error in
    // Only AccountNotFoundError
}
.catch { error in
    // All other errors
}
```

I don't think it's implementable for API users in an extension because there's a need to stop the promise chain, which I don't think can be done without a `throw` with the current codebase.

BennX:
I think the basic idea here would be to add a new catch method to the State which takes an NSError. Alternatively an Error & Compareable would be possible:
```Swift
final func `catch`(on q: DispatchQueue, policy: CatchPolicy, `catch`: NSError, else resolve: @escaping (Resolution<T>) -> Void, execute body: @escaping (Error) throws -> Void) {
    pipe { resolution in
        switch (resolution, policy) {
        case(let .rejected(error, token), _) where (error as NSError) == `catch`:
            contain_zalgo(q, rejecter: resolve) {
                token.consumed = true
                try body(error)
            }
        default:
            resolve(resolution)
        }
    }
}
```
Inside the regular `catch` it is needed to have a check if the error is already consumed. If so do not execute the body.

```Swift
case (let .rejected(error, token), _):
    contain_zalgo(q, rejecter: resolve) {
        if token.consumed {
            resolve(resolution)
        } else {
            token.consumed = true
            try body(error)
        }
    }
}
``` 

Now the Promis itself need the function to catch with an NSError:
```Swift
@discardableResult
public func `catch`(on q: DispatchQueue = .default, policy: CatchPolicy = .allErrorsExceptCancellation, _ `catch`: NSError, execute body: @escaping (Error) -> Void) -> Promise {
    state.catch(on: q, policy: policy, catch: `catch`, else: { _ in }, execute: body)
    return self
}
```

Catching then can look like this:
```Swift
func testCatchWithSpecificError() {
    let ex = expectation(description: "")
    
    Promise<()>(value: ()).then {
        throw Error.more as NSError
    }.catch(Error.dummy as NSError) { _ in
        XCTFail()
    }.catch(Error.more as NSError) { _ in
        ex.fulfill()
[...]

PR #960, opened 2018-11-20
[Proposal] Provide for separate handling of specific errors in promise chains.

nathanhosselton:
## Intro
Currently all errors in a promise chain propagate down to a single `catch` handler at the end. Therein, the identity of the error must often be reasoned out with comparative statements and then handled according to its identity. This can lead to otherwise clean chains becoming bottom-heavy with logic in the `catch`.

Swift's do-catch provides for error pattern matching in its catch handlers, raising the error identification logic up one level. This PR provides an implementation to introduce this pattern into PromiseKit, allowing for separate catch handlers for specific errors.

_refs #639_

## Usage
The new `catchOnly` method included in this PR works the same as the current `catch` method, except that it takes a required `Error` parameter, and only executes its handler if the promise chain rejects with that error.

```swift
firstly {
    API.findUser(named: "Taylor")
}.done { user in
    //…
}.catchOnly(API.Error.noResults) {
    //Let the user know there's no results for their query
}.catch { error in
    //Handle other errors like timeouts
}
```

In the above, the `catchOnly` handler would only execute if the API request rejected with a `noResults` error. The final `catch` handler would then no longer execute since the error was already handled. Likewise, if some other error occurred, the `catchOnly` handler would be skipped and only the final `catch` handler would be executed.

An arbitrary number of `catchOnly` calls may be chained, but the chain must still be terminated with a final `catch`. An overload for `catchOnly` which takes a type instead of an object is also provided for matching on a family of errors.

## Implementation
The new functions work similarly to the current `catch` function, only instead of returning a `PMKFinalizer` they return a void promise which is fulfilled if the error is handled (or if the promise was fulfilled), and rejected with the unhandled error otherwise to allow it to propagate to successive catch handlers.

Since no changes were made to existing API or implementations, a major version rev should not be necessary. I have included some trivial tests in the PR as evidence that the new functions behave as intended, but I am happy to flesh these out or strip them entirely upon request.

I've not included any implementation for the Objective-C interface (not sure if it's necessary or even desired). But I am happy to do so if requested.

Ideally, I would have preferred a simple `catch(Error.someError)` signature both for succinctness and keeping in line with Swift's do-catch, but it surfaced the usual compiler confusion that PMK's overloads have dealt with in the past. I settled on `catchOnly` to get around this, but I actually rather like it now, and its visual distinctness in code from the base `catch` is probably for the best. Still, open to name changes, of course.

## Limitations
Since the new `catchOnly` functions return a promise, it is possible to chain off of them with functions other than `catchOnly` and `catch`. This isn't desirable, but is also unlikely to to be utilized, I believe, since the underlying value from the promise would have been consumed by `catchOnly`. A chained `then` or `done` for instance would have no context. I had originally intended to return a `PMKFinalizer` to avoid this, but that solution quickly proved to have too many issues.

Error objects with associated values, while technically supported, make little sense in the usage of `catchOnly` since concrete associated values would need to be supplied in order to pattern match. However, the type-accepting overload of `catchOnly` can still prove useful in these cases.

Errors passed in must conform to `Equatable`, but this is given for free in most situations in Swift 4 now.

## Alternatives
`recover` may already be used in a similar manner to `catchOnly` (as described in #639). And this doesn't appear to be a hotly requested or needed feature by any means. So, leaving it out of PromiseKit probably isn't going to cause any friction.

### DIFF
--- a/Sources/Catchable.swift
+++ b/Sources/Catchable.swift
@@ -54,11 +54,11 @@ public extension CatchMixin {
      - Note: Since this method handles only specific errors, supplying a `CatchPolicy` is unsupported. You can instead specify e.g. your cancellable error.
      - SeeAlso: [Cancellation](http://promisekit.org/docs/)
      */
-    func catchOnly<Error: Swift.Error>(_ only: Error, on: DispatchQueue? = conf.Q.return, flags: DispatchWorkItemFlags? = nil, _ body: @escaping() -> Void) -> PMKCascadingFinalizer where Error: Equatable {
+    func catchOnly<E: Swift.Error>(_ only: E, on: DispatchQueue? = conf.Q.return, flags: DispatchWorkItemFlags? = nil, _ body: @escaping() -> Void) -> PMKCascadingFinalizer where E: Equatable {
         let finalizer = PMKCascadingFinalizer()
         pipe {
             switch $0 {
-            case .rejected(let error as Error) where error == only:
+            case .rejected(let error as E) where error == only:
                 on.async(flags: flags) {
                     body()
                     finalizer.pending.resolver.fulfill(())
@@ -85,11 +85,11 @@ public extension CatchMixin {
      - Parameter execute: The handler to execute if this promise is rejected.
      - SeeAlso: [Cancellation](http://promisekit.org/docs/)
      */
-    func catchOnly<Error: Swift.Error>(_ only: Error.Type, on: DispatchQueue? = conf.Q.return, flags: DispatchWorkItemFlags? = nil, policy: CatchPolicy = conf.catchPolicy, _ body: @escaping(Error) -> Void) -> PMKCascadingFinalizer {
+    func catchOnly<E: Swift.Error>(_ only: E.Type, on: DispatchQueue? = conf.Q.return, flags: DispatchWorkItemFlags? = nil, policy: CatchPolicy = conf.catchPolicy, _ body: @escaping(E) -> Void) -> PMKCascadingFinalizer {
         let finalizer = PMKCascadingFinalizer()
         pipe {
             switch $0 {
-            case .rejected(let error as Error):
+            case .rejected(let error as E):
                 guard policy == .allErrors || !error.isCancelled else {
                     return finalizer.pending.resolver.reject(error)
                 }
@@ -156,7 +156,7 @@ public class PMKCascadingFinalizer {
      - Note: Since this method handles only specific errors, supplying a `CatchPolicy` is unsupported. You can instead specify e.g. your cancellable error.
      - SeeAlso: [Cancellation](http://promisekit.org/docs/)
      */
-    public func catchOnly<Error: Swift.Error>(_ only: Error, on: DispatchQueue? = conf.Q.return, flags: DispatchWorkItemFlags? = nil, _ body: @escaping() -> Void) -> PMKCascadingFinalizer where Error: Equatable {
+    public func catchOnly<E: Swift.Error>(_ only: E, on: DispatchQueue? = conf.Q.return, flags: DispatchWorkItemFlags? = nil, _ body: @escaping() -> Void) -> PMKCascadingFinalizer where E: Equatable {
         return pending.promise.catchOnly(only, on: on, flags: flags) {
             body()
         }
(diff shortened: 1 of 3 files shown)


mxcl:
Is this worth it?

You don't usually care which error happens, and when you do, you often enough need associated values. Most of the time with Promises when I care about specific errors it is in order to recover. Which this PR doesn't handle.

The fact that our `catch` cannot be as elegant as a Swift `catch` means this PR could never deliver the kind of value everyone *really* wants.

However it's not as though I've never wanted this. It's just that most of the times I've wanted it I've got a good Swift error with associated values. Seems like this could be mitigated by provided the error to the handler, but since it will be passed by type, you'd have to unwrap it to get that associated value still.

Either way returning a Promise is a definite no. I've seen in the past when `catch` returned Promise people getting confused about the chains, this becomes much easier when chains are returned via multiple levels of functions, which is typical with promises, you lose track of where the error is handled leading to code that expects errors but isn't getting them because an earlier `catch` already consumed the error.

We would have to return a PMKFinalizer which is adapted to handle subsequent catches.

nathanhosselton:
So even though this may not be merged I wanted to go ahead and push up what some of the discussed changes/issues might look like implemented/addressed. Both for fun and reference.

- I created a new `PMKCascadingFinalizer` object to be returned by the `catchOnly` methods. It behaves the same as `PMKFinalizer` only wraps a promise instead of a guarantee. It has both `catchOnly` methods as well as `catch` which call their respective `Promise<T>` counterparts for chaining and chain completion.
- I implemented `recoverOnly` with both an object and type accepting variant to mirror `catchOnly`.
- The type accepting variants of `catchOnly` and `recoverOnly` can have a catch policy specified for handling or ignoring any cancellation errors within the passed type.

If there ends up being interest to merge this just lmk if there are any changes desired and I'll happily oblige.

mxcl:
I will merge this if we can do the same as a Swift `catch` statement. I believe this means:

```swift
do {
    //…
} catch ErrorType {
    //…
} catch ErrorType.specificError {
    //…
}
```

Anything else?

nathanhosselton:
AFAIK this isn't possible. So feel free to close.

mxcl:
Seems like we can do both, would look like:

```
foo.then {
    //…
}.catch(ErrorType.self) {
    //…
}.catch(ErrorType.specificCase) {
    //…
}
```

nathanhosselton:
Forgive me, I don't think I'm totally following. The last example feels like what is currently implemented (only using an overload). What would you like to see on top of this?

mxcl:
Your usage example did not demonstrate this:

```swift
}.catch(ErrorType.self) {
    //…
}
```

So you’re saying this is also supported?

nathanhosselton:
Oh yep. Apologies for omitting that use case in the example, but it's there, including falling back to the (still required) final `catch` handler if prior specific handlers are not triggered.

mxcl:
K finally, which of these is triggered?

```swift
firstly {
    throw ErrorType.specificCase
}.catch(ErrorType.self) {
    //…
}.catch(ErrorType.specificCase) {
    //…
}.catch {
    //…
}
```

nathanhosselton:
Let me verify. I may not have handled this, meaning the first would be triggered with the middle skipped, even though it's a more specific match. I'm assuming you would like the middle catch to be called in this case.

nathanhosselton:
Yep, the first catch would be called with the following two skipped. I have a test verifying this.

mxcl:
> I'm assuming you would like the middle catch to be called in this case.

Well the order is bad, so the first should be called and not the second I think. For example with this:

```swift
enum E: Error {
    case foo
}

do {
    throw E.foo
} catch is CocoaError {
    print(0)
} catch E.foo {
    print(1)
} catch E.foo {
    print(1.4)
} catch is E {
    print(2)
} catch {
    print(3)
}
```

`1` is printed.

nathanhosselton:
Fair enough. So, it'd be considered a developer error in that case.

For quick reference, here are all of the features currently implemented in this PR:

```swift
firstly {
    throw ErrorType.specificCase
}.recoverOnly(ErrorType.specificCase) {
    //…
}.recoverOnly(ErrorType.self) { error in
    //…
}.recover {
    //Executed only if specific recovers are not triggered
}.catchOnly(ErrorType.specificCase) {
    //…
}.catchOnly(ErrorType.self) { error in
    //…
}.catch {
    //Executed only if specific catches are not triggered
}
```

mxcl:
Cool, well this is good IMO. I'd like to review the tests to ensure everything is covered. Also I'd prefer just `catch` not `catchOnly` (etc.) what you think that?

mxcl:
Even if the parameter had no default?

nathanhosselton:
Okay I played around again and remember the issue. It's the two new variants that don't disambiguate well between each other (`Error.specificError` vs `Error.self`). Whenever only one is named `catch` it's fine, but once both are Swift starts getting confused.

It actually does compile out now once the closure is completed. I assume because I have since removed the closure's parameter in one and not the other. But up until that point Swift generates false errors (e.g. `Missing parameter on: in call…`)

nathanhosselton:
Okay I don't think this is bad at all now actually. I think they can all be `catch`/`recover` and it'll be good so long as the placeholder autocompletions are used or the closure parameters aren't otherwise omitted.

nathanhosselton:
One case where this could be confusing is when switching between the two variants after completing the code. It'll be easy to forget to remove/include the closure parameter for the error object and Swift won't provide the correct compiler error.

This is unrelated to dropping "Only" off of the function names. Just something that's occurring to me now.

drekka:
I like this. I was thinking it would be really useful to have this `.catchOnly(MyError.self) {` as I might handle situations where data errors are handled differently to networking errors and it would be nice for PromiseKit to save me from my obsession with `switch` statements (Only a slight exaggeration :-)

GarthSnyder:
I have some code that will use this once it gets into the v7 beta. The design looks good to me.

mxcl:
Can we change to a v7 submission @nathanhosselton

nathanhosselton:
@GarthSnyder @mxcl new PR onto v7 at #1024

mxcl:
Closing in favor of #1024

PR #961, opened 2018-11-22
Added facilities for controlling console output.

neallester:
I did not test the case where a pending promise is deallocated. I ran out of time trying to figure out how to trigger it.

### DIFF
--- a/Documentation/FAQ.md
+++ b/Documentation/FAQ.md
@@ -370,6 +370,34 @@ Kitura.addHTTPServer(onPort: 8888, with: router)
 Kitura.run()
 ```
 
+## How to I control console output?
+
+By default PromiseKit emits console messages when certain events occur.  These events include:
+
+```swift
+public enum LogEvent {
+    case waitOnMainThread
+    case pendingPromiseDeallocated
+    case cauterized(Error)
+}
+
+```
+
+You may turn off console logging with:
+
+```swift
+PromiseKit.loggingPolicy = .none
+```
+
+You may redirect the output to your own custom logger with
+
+```swift
+let loggingClosure: (PromiseKit.LogEvent) -> () = { event in
+    // Log event
+}
+PromiseKit.loggingPolicy = .custom(loggingClosure)
+```
+
 ## My question was not answered
 
 [Please open a ticket](https://github.com/mxcl/PromiseKit/issues/new).
--- a/Documentation/FAQ.md
+++ b/Documentation/FAQ.md
@@ -373,29 +373,14 @@ Kitura.run()
 ## How do I control console output?
 
 By default PromiseKit emits console messages when certain events occur.  These events include:
+- A promise or guarantee has blocked the main thread
+- A promise has been deallocated without being fulfilled
+- An error which occurred while fulfilling a promise was swallowed using cauterize
+
+You may turn off or redirect this output by setting a thread safe closure in [PMKCOnfiguration](https://github.com/mxcl/PromiseKit/blob/master/Sources/Configuration.swift) **before** processing any promises. For example, to turn off console output:
 
 ```swift
-public enum LogEvent {
-    case waitOnMainThread
-    case pendingPromiseDeallocated
-    case cauterized(Error)
-}
-
-```
-
-You may turn off console logging with:
-
-```swift
-PromiseKit.conf.loggingPolicy = .none
-```
-
-You may redirect the output to your own custom logger with
-
-```swift
-let loggingClosure: (PromiseKit.LogEvent) -> () = { event in
-    // Log event (must be thread safe)
-}
-PromiseKit.conf.loggingPolicy = .custom(loggingClosure)
+conf.loggingClosure = { event in }
 ```
 
 ## My question was not answered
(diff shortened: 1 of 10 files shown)


mxcl:
Seems good. Can we add this to the already existing configuration object?

neallester:
Sure, should I close this pull request and resubmit when that's done?

mxcl:
No need you can pick she the amendment here

neallester:
One other thing I didn't do is add the new Logging test to AllTests (I wasn't sure exactly how to do it or if that was even the right place) so the new test may not run on Linux.

mxcl:
I tried to fix the CI errors, but I don’t have push access to your branch, so you’ll have to do it.

The `Logging.swift` file is not in `Sources` so CocoaPods does not pick it up.

neallester:
@mxcl, you're now a collaborator on the fork. Are you suggesting I move the file Logging.swift to the directory Sources.swift?

mxcl:
> you're now a collaborator on the fork

Thanks, but well, there's some setting you can set so when you PR the destination user of the PR can push to your repo for that branch. I forget where it is.

mxcl:
I think we can simplify this to just:

```swift
public struct PMKConfiguration {
    var log: ((PMKLogEvent) -> Void)? = {
        print($0)
    }
}
```

This makes the API less overwhelming to users, uses Swift standards for “nothingness” (ie. `nil`) and allows the user to do whatever they want while providing a sensible default.

What do you think?

mxcl:
And we can drop ` waitOnLogging()` and the `loggingQueue`.

neallester:
I originally thought I would write at least two tests, but when I finished
the first set of assertions I realized they left things in a convenient
(and non-default) configuration for the second set of tests (those after
the comment "
*Verify waiting on main thread in Promise is logged*". The existing tests
could be broken into two tests or even four. There would just be some
duplicate set up between three of the tests (I suppose these could be
broken into a separate test class to refactor out the common setup but that
seems like overkill for the simple setup and adds some state between the
tests).

mxcl:
As it stands the tests are failing on CI and it's pretty hard to debug them since it is 30 lines of code which I cannot easily reason about.

neallester:
> And we can drop ` waitOnLogging()` and the `loggingQueue`.

Really? Isn't there a risk of a race between one thread changing the logging closure and another thread firing that closure?

mxcl:
> Really? Isn't there a risk of a race between one thread changing the logging closure and another thread firing that closure?

The rules for `PMKConfiguration` are that you must set these things before any Promise machinery executes. I used to have enforcement for this using `dispatch_once` but Swift does not expose these methods sadly, so when I was making PromiseKit usable for pure swift, this is the result: a contract to consumers.

Libraries that use PromiseKit could change these things, but that would make them bad libraries.

neallester:
Would you object to addressing these four failing assertions with

#if swift(>=4.2)

mxcl:
We test for all the swifts we support to insure that we never break on all the Swifts we support, so yeah, sorry.

neallester:
I think the problem is that using an enum with an associated value within a string interpolation gave a different output in previous versions.

mxcl:
> I think the problem is that using an enum with an associated value within a string interpolation gave a different output in previous versions.

Well, you could ifdef for each version.

Still, I suggested dropping the enum above anyway.

neallester:
This change should work and avoids ifdef by using a switch.

mxcl:
How do you feel about this just being a closure on `PMKConfiguration` without the rest, I believe in keeping APIs minimal.

neallester:
> How do you feel about this just being a closure on PMKConfiguration without the rest, I believe in keeping APIs minimal.

There is a very minor drawback I've detailed below. I'm ok with:

- Remove LoggingPolicy enum
- Remove PMKConfiguration.loggingPolicy
- Remove PMKConfiguration.logConsoleClosure
- Rename activeLoggingClosure to loggingClosure and making it public. What is now the implementation of logConsoleClosure will become the default value for loggingClosure.
- Remove Logging.log
- Keep public LogEvent enum

One purpose of LoggingPolicy was to make it easier to turn logging back on after turning it off. Since that isn't a supported use case I agree it doesn't make things significantly easier than just assigning a closure to conf.activeLoggingClosure (renamed conf.loggingClosure).

The only disadvantage of doing this is it would enable code from outside of PromiseKit to send LoggingEvents to the closure. While that isn't an intended use case, I don't see how adding the capability (as a side effect of the simplification you are proposing) is so terrible.

I agree that the log feature is a useless indirection and should be removed. I created it as part of the public API and then made it internal when I realized it was dumb to allow code outside of PromiseKit to trigger log events without stopping to think that switching it to internal eliminated the reason I had it there in the first place.

I think we want to keep LogEvent as a public enum for the parameter of the logging closure. If you discover some other event which needs logging adding it as an enum case will force anyone with a custom logging closure to address that new case in their closure (which might be different than for the existing events)

Is this what you had in mind?.

mxcl:
Will be live when this finishes: https://travis-ci.org/mxcl/PromiseKit/builds/463580751

I changed the function name, hope that's cool. See the recent commits to master for changes.

neallester:
Of course that's fine, but would you please update the name at the bottom
of FAQ.md

Neal

mxcl:
Oops, thanks.

mxcl:
Done.

PR #963, opened 2018-11-29
Create void fulfill function

LarsStegman:
Hello everybody

I find it annoying to have to add a Void value every time I need to fulfill a Void promise. It can also be confusing how to fulfill Void promises for people who are not familiar with how to instantiate a Void value.

This PR would allow users to replace this syntax
```
resolver.fulfill(())
```

with 

```
resolver.fulfill()
```

Thanks!

### DIFF
--- a/Sources/Resolver.swift
+++ b/Sources/Resolver.swift
@@ -65,6 +65,11 @@ extension Resolver where T == Void {
             fulfill(())
         }
     }
+
+    // Fulfills the promise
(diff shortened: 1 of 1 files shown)


mxcl:
Many thanks 👍🏻 I'm surprised we didn't already add this… much like yourself!

mxcl:
Unfortunately we are having to revert this until PMK7, see #990.

PR #973, opened 2018-12-13
Log when unfulfilled guarantee is deallocated

neallester:
Now logs when a Guarantee is deallocated with being fulfilled.

Fixed bug where pendingPromiseDeallocated event was incorrectly reported as a waitOnMainThread event. Added test which should have caught this bug in the first place.

This is a breaking change because it adds a case to the LogEvent enum.


neallester:
Travis CI build [1731](https://travis-ci.org/mxcl/PromiseKit/builds/467761085) failed at step 7, but Travis CI build [1729](https://travis-ci.org/mxcl/PromiseKit/builds/467698921) succeeded at step 7 even though there was no code change between the two builds (only an empty commit to kick off another CI build.

The CI failure appears to be a Travis issue.

mxcl:
Sorry, this slipped away from me and I have no system for pinging me.

mxcl:
Well, this is completely broken, so I will have to revert.

It reports *all* Guarantees as deallocated pending.

mxcl:
I will from now on be far firmer with my declaration that this library is not receiving new features :( I haven't the time to test things.

mxcl:
K fortunately I have found a simple enough workaround: #1002

neallester:
Sorry I missed the implication of the Box indirection although it's hard to feel too guilty about it. Since unfulfilled promises already emit console output/log on deallocation doing so for Guarantees looks to me more like completing an existing feature than adding a new feature. At the moment I don't expect to use Guarantees so I submitted the PR for the sake of completeness when I happened to notice unfulfilled Guarantees went to their doom silently.

Is there anything you need from me?

PR #975, opened 2018-12-18
Revised dispatching code ported to v7

GarthSnyder:
Fixes #888.


mxcl:
Are the `.travis.yml` changes required for this PR?

GarthSnyder:
Yes. See comments [here](https://github.com/mxcl/PromiseKit/pull/888#issuecomment-448116582) in the thread for #888. A relatively recent compiler is required, even for Swift 3, so the test configuration has to change.

This `.travis.yml` also has some support for harvesting build artifacts (for Travis debugging) and auto-building arbitrary branches. Those features have no effect if you don't define Travis environment variables to enable/configure them. But if you don't want to accept those changes, you can just reverse commits 0b936fd, f78eaf1, and 0da820a. (Or I can prepare a branch with those changes.)

mxcl:
K I see, is it possible to lose all the rearrangements? It's hard for me to reason about the changes in the travis file because of this. And I'm nervous about the changes there because I have spent gosh knows how many hours on it and it's fiddly and flakey and breaks mean that are CI testing is less certain.

GarthSnyder:
Don't I know it! It was something of a project to get the Travis configuration updated.

I put the order back to mirror the trunk version, but it still doesn't diff very cleanly because a lot of sections are self-similar. So here's a guide. General comments: 

1) in some of the existing configurations, an xcode10.1 image was used but the name identified it as Xcode 10.0. I updated the names to match the actual image version.

2) I updated some of the names to make them more detailed/explicit so that it was easier to tell exactly what was failing in the Travis dashboard.

3) The main reason to touch this file at all is that you need a current compiler for Swift 3, so some cases that addressed Swift 3 in combination with Xcode < 10 or tools < 4.2 were removed. Swift 4 works back to Xcode 9 and 4.0.3, so those configurations stayed.

4) The original Linux schedule seems to have compiled for Swift 3 under 4.0.3, 4.1.2, and 4.2.1, but only 4.1.2 for Swift 4. And it tested only 4.0.3/Swift 4 and 4.2.1/Swift 4. This seemed a bit odd. It now builds for Swift 3 on 4.2.1 and for Swift 4 on 4.0.3, 4.1.2, and 4.2.1, with testing being done on these same combinations. I can put this all back if I've messed something up. 

5) Should there be Swift 4 tests for iOS and tvOS? These seem to be built for Swift 3 only.

Detailed list of changes:

```
carthage/compile rm xcode8.3
carthage/compile rm xcode9.2
carthage/compile rm xcode9.4
carthage/compile add xcode10.1 and Swift 3.4 combination
carthage/compile make xcode10.1 and Swift 4 request explicit SWIFT_VERSION=4.2

pod/lint make baseline xcode10.1, SWIFT=3.4 instead of xcode8.3/Swift 3.1
pod/lint rm xcode8.3 for Swift 3
pod/lint rm xcode9.2 for Swift 3
pod/lint rm xcode9.4 for Swift 3

linux/compile make baseline tools 4.2.1 instead of tools 4.0.3 (Swift 3)
Other changes as outlined above

mac/test make baseline xcode10.1/3 instead of xcode8.3/3
mac/test rm xcode9.2/4
mac/test rm xcode9.4/4
mac/test fix name of xcode10.1/4 scenario to reflect Xcode 10.1

ios/test fix name of xcode10.1/3 scenario to reflect Xcode 10.1 

tvos/test fix name of xcode10.1/3 scenario to reflect Xcode 10.1
```

mxcl:
I have pushed the start of `v7`, and I have stripped it quite a bit, my thinking being that v7 can be modernized, and we'll maintain v6 as the back-compat version.

This means we have a lot less worry about older versions having ambiguity for this patch.

Sorry that the branch now needs merging again. If you like, I can do it.

GarthSnyder:
There doesn’t seem to be an Xcode project at the current V7 tip. Is this the intended state?

mxcl:
> There doesn’t seem to be an Xcode project at the current V7 tip. Is this the intended state?

Yes, you have to generate it: `swift package generate-xcodeproj`

mxcl:
Don’t worry about CI, it's not going to work until Travis have an Xcode 10.2 image.

mxcl:
Looks ready to merge to me, what you say?

GarthSnyder:
Yes, OK to merge if you're ready for it.

There are a couple of issues (that we already talked about) yet to address, so I want to be sure you're aware that this is still WIP code from my perspective. (Though it should be 100% working and near-final API.) If you'd rather defer until I have something that looks completely done to me, that would be fine too. I mostly just wanted to get synced with the v7 line and get a sense of the v7 changes before finalizing.

PR #983, opened 2019-01-15
Start running CI against Swift 5 snapshots

mxcl:


mxcl:
Not sure why this doesn’t work, putting it on hold for now.

PR #986, opened 2019-01-17
Use PackageDescription API v4 in Installation.md

VFUC:
Updates Installation.md to use PackageDescription API v4

Also uses Swift syntax highlighting instead of ruby


mxcl:
Thanks 👍🏻

PR #987, opened 2019-01-18
Adding more queue async functions.

drekka:
# Additional extension functions for queuing initial promises 

## The problem

I was attempting to create a new promise on a background queue and felt the way I had to do it wasn't as easy or as nice as I'd like it to be. My problem was that I needed it to execute some code that contained more asynchronous functionality.  So I ended up with something like this (very simplified example):

```swift
return Promise<Int> { seal in
    backgroundQueue.async {
         seal.fulfill(5)
    }
}
```

After looking at the PromiseKit async extensions I had a few things about it I wanted to change:

* Even though they could produce both promises and guarantees, there was nothing in the signature to make it clear to other developers. ie:

    ```swift
    // Guarantee
    backgroundQueue.async(.promise) { () -> Int in
        return 5
    }

    // Promise
    backgroundQueue.async(.promise) { () -> Int in
        return 5
    }.catch { error in
        //...
    }
    ```
* Apart from the `catch` closure there's no way to tell whether it's a promise or guarantee coming back.
* Even when returning a guarantee, the function 'asks' for a promise with the `.promise` argument. 
* As per the code above, the `async(...)` function could not pass a resolver for sealing.

## Proposed solution in this PR

This PR contains code that does this:

* Adds new async methods with these signatures:
 
    ```swift
    func asyncPromise<T>(_ ofType: T.Type, ..., execute body: @escaping () throws -> T) -> Promise<T>
    final func asyncPromise<T>(_ ofType:T.Type, ..., resolver body: @escaping (Resolver<T>) -> Void) -> Promise<T>
    func asyncGuarantee<T>(_ ofType: T.Type, ..., execute body: @escaping () throws -> T) -> Guarantee<T>
    final func asyncGuarantee<T>(_ ofType:T.Type, ..., resolver body: @escaping (Resolver<T>) -> Void) -> Guarantee<T>
    ```

This lets you write code like this:

```swift
// Guarantee
backgroundQueue.asyncGuarantee(Int.self) {
    return 5
}
backgroundQueue.asyncGuarantee(Int.self) { seal in
    seal.fulfill(5)
}

// Promise
backgroundQueue.asyncPromise(Int.self) {
    return 5
}.catch { error in
    //...
}
backgroundQueue.asyncPromise(Int.self) { seal in
    seal.fulfill(5)
}.catch { error in
    //...
}
```

Please see the unit tests for working examples. The advantages of these options (as I see them) are:

* Each call is as close as I can get it to declaring a normal promise or guarantee. ie. `q.asyncPromise(Int.self) {...` reads like `Promise<Int> {...`.
* It's clear whether you are creating a promise or guarantee. I also tried a style using the established `.async(.promise) {` style to add the sealable methods, but didn't like they way they turned out.
* The closure's no longer need to have argument types declared. They are now inferred.

Please feel free to get back to me with ideas and comments.

Thanks
Derek

### DIFF
--- a/Documentation/Appendix.md
+++ b/Documentation/Appendix.md
@@ -168,30 +168,65 @@ will deadlock.
 can be found in the ticket tracker.
 
 So, if you want to start a chain by dispatching to the background, you have to use
-`DispatchQueue.async`:
+`DispatchQueue.async` like this for guarantees:
 
 ```swift
-DispatchQueue.global().async(.promise) {
-    return value  
+DispatchQueue.global().async(.promise) { () -> String in
+    return "abc"  
 }.done { value in
     //…
 }
 ```
 
-However, this function cannot return a promise because of Swift compiler ambiguity
-issues. Thus, if you must start a promise on a background queue, you need to
-do something like this:
-
+And this to return promises:
 
 ```swift
-Promise { seal in
-    DispatchQueue.global().async {
-        seal(value)
-    }  
+DispatchQueue.global().async(.promise) { () -> String in
+    return "abc"  
+}.done { value in
+    //…
+}.catch { error in
+    //… 
+}
+```
+
+Alternatively there are a set of extensions that are more explicity and don't require you to add the closure argumnet types. They look like this:
+
+```swift
+DispatchQueue.global().asyncGuarantee(String.self) {
+    return "abc"  
 }.done { value in
     //…
 }
+
+DispatchQueue.global().asyncPromise(String.self) {
+    return "abc"  
+}.done { value in
+   //…
+}.catch { error in
+   //… 
+}
 ```
+
+All these four functions work well as long as your closures don't contain asynchronous code. If they do, then you'll need to use these varients which pass in resolvers:
+
+```swift
+DispatchQueue.global().asyncGuarantee(String.self) { seal in
+    seal.fulfill("abc")  
+}.done { value in
+    //…
+}
+
+DispatchQueue.global().asyncPromise(String.self) { seal in
+    seal.fulfill("abc")  
+}.done { value in
+    //…
+}.catch { error in
+    //… 
+}
+```
+
+
 
 Or more simply (though with caveats; see the documentation for `wait`):
 
--- a/Sources/Guarantee.swift
+++ b/Sources/Guarantee.swift
@@ -165,10 +165,11 @@ public extension Guarantee where T == Void {
 
 
 public extension DispatchQueue {
+
     /**
      Asynchronously executes the provided closure on a dispatch queue.
 
-         DispatchQueue.global().async(.promise) {
+         DispatchQueue.global().async(.promise) { () -> Int in
              md5(input)
          }.done { md5 in
              //…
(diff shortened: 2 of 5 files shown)


mxcl:
Basically, -1 from me. Sorry.

This would make these functions inconsistent with the rest of our library, of which including all the extensions is a very large API surface now.

I agree with the sentiment that it is useful to know if a PromiseKit function returns a guarantee or a promise without having to read its documentation, but unfortunately we have a lot of cases like this, and I'm in general unwilling to uglify the library with naming them promise or guarantee the whole time. 

> The `(.promise, )` naming scheme was added only because Swift makes disambiguating the functions tedious otherwise, it is a namespace and not a statement about the return type. While I acknowledge this is confusing, and would change it to something else now if I could, we do not use it consistently so even if we added `.guarantee` to the enum we would only be annotating a small portion of our APIs that return Promises/Guarantees.

Adding the explicit type as a parameter is not consistent with the rest of our library and isn't necessary IMO since you can add the return type to the closure. Swift does not provide eg. `map(Int.self)` the only occasions where the standard library do this are for non-closure functions like `JSONDecoder.decode` since otherwise you would have to specify the type on the variable declaration.

Finally I'd argue if it isn't clear if you have a Promise or a Guarantee in specific code portions that maybe it's a code smell? Like for me, it's usually clear, since the code is inside a function that returns a chain that is one or the other. Certainly it is possible to have chains that accidentally change from guarantees to promises, and having more clarity would help, but as said 95% of our APIs that return Guarantees are not the ones in this PR, so we're not really helping overall.

Ultimately though you know if you have a Guarantee chain or a Promise chain, does it demand a `catch`? If so, it's a promise chain.

I did once consider making Guarantee/Promise initializers that start on another queue, but in the end decided it was not easy to make it clear it was another queue executing. Adding an extension to `DispatchQueue` means we are using the user's existing knowledge about GCD to communicate what is happening.

drekka:
Yep. I initially played around with adding `.guarantee` to the namespace, but that didn't stop me from saying `.promise` and returning a guarantee which I thought was disturbing and would be a confusing issues for developers who are not familiar with the API. 

I also messed with just using `.async{`, but I liked the idea of developers not having to specify the closure argument types. 

I agree with you on the type argument, not the best but I was trying for readability.

I'll re-engineer back to `.async` in line with your comments as I don't mind either option. The thing I mostly needed was the sealable option.

drekka:
Hmmm. Need some advise. In the test code all the tests for the async functions need the closure argument to be fully specified. ie.

```swift
backgroundQueue.async(.promise) { () -> Int in
    return 5
}
```

Some of the comments leave the closure argument specification out, but in my tests that won't compile. ie.

```swift
backgroundQueue.async(.promise) { // Won't compile as it cannot infer type!
    return 5
}
```

Which lead me to believe the `.promise` argument actually wasn't needed. So I tested removing it and the following code compiled just fine:

```swift
backgroundQueue.async { () -> Int in
    return 5
}

backgroundQueue.async { () -> Int in
    return 5
}.catch { _ in }

// and with the seal code I'm adding:
backgroundQueue.async { (seal: Resolver<Int>) in
    return 5
}

backgroundQueue.async { (seal: Resolver<Int>) in
    return 5
}.catch { _ in }
```

Am I missing something?

mxcl:
> Am I missing something?

It's now difficult for people to use the version that Apple provides that returns neither Promise or Guarantee.

drekka:
Hi, can you give me some more details on what the confusion would be? 

I added a couple of test functions to call Apple's `.async {....}` functions and had no problem with code completion, etc. The only thing I noticed was that code completion listed the PK functions with the GCD ones. Is that what you mean? :-)

drekka:
Hmm. been thinking about this. I think I'll do it as an seperate library instead.

Issue #990, opened 2019-01-23
"Ambiguous use of fulfill" when upgrading to 6.7.1


michaelkirk-signal:
* Please specify the PromiseKit major version you are using

PromiseKit (6.7.1) installed via cocoapods, recently upgraded from 6.5.3, but I believe the problem was introduced with 33e7eba in 6.6.0

**description**

We have a lot of methods which wrap legacy success/failure style methods with a promise like this:

```
 private func registerForTextSecure(verificationCode: String,
                                       pin: String?) -> Promise<Void> {
        return Promise { resolver in
            tsAccountManager.verifyAccount(withCode: verificationCode,
                                           pin: pin,
                                           success: resolver.fulfill,
                                           failure: resolver.reject)
        }
    }
```

Upon upgrading from 6.5.3 to 6.7.1 we get the compiler error: `Ambiguous use of fulfill`

The following patch resolves the issue:

```
$ git diff
diff --git a/PromiseKit/Sources/Resolver.swift b/PromiseKit/Sources/Resolver.swift
index 297de03..be89de0 100644
--- a/PromiseKit/Sources/Resolver.swift
+++ b/PromiseKit/Sources/Resolver.swift
@@ -62,14 +62,9 @@ extension Resolver where T == Void {
         if let error = error {
             reject(error)
         } else {
-            fulfill()
+            fulfill(())
         }
     }
-
-    /// Fulfills the promise
-    public func fulfill() {
-        self.fulfill(())
-    }
 }
 #endif
 ```

So it seems like this was introduced with 33e7eba in 6.6.0. (A change which I'm sympathetic to by the way!)

Is there a different way we should be wrapping these handlers?

mxcl:
Thanks, this is a serious API breakage and I thank you for your patience. I am reverting the change even though it has been months and no doubt some people have worked-around this issue. This is unfortunate, but it’s the right thing to do.

We will reintegrate this for PMK7 and the ambiguity will then have to stand, the workaround will be to pass a closure that calls `fulfill()`.

michaelkirk-signal:
Thanks for the prompt response!

mxcl:
Happy to help! Though it’s only possible to be this prompt because I'm full-time.

Donate if you can! https://patreon.com/mxcl

michaelkirk-signal:
> Happy to help! Though it’s only possible to be this prompt because I'm full-time.
> 
> Donate if you can! https://patreon.com/mxcl

I'm already on it. =)

[...]

PR #992, opened 2019-01-23
All changes should do a full lint

mxcl:
For example, we did not verify that a recent change would not affect compilation of all extensions which may have caused a bug: #990


mxcl:
Actually, no, we do a full lint when we push, so this is just a waste of resources.

PR #998, opened 2019-01-25
Make umbrella header import all public .h files with proper public import syntax

dfed:
Umbrella headers need to import all public Objective-C header files. Without this, projects may get compiler warnings like: `/<module-includes>:1:1: Umbrella header for module 'PromiseKit' does not include header 'PMKCoreLocation.h'`

This PR adds all public header files to the PromiseKit umbrella header. It also changes the import syntax to use `#import <PromiseKit/PublicHeader.h>`, which is the proper way to import module-specific files in a public header.

### DIFF
--- a/Sources/PromiseKit.h
+++ b/Sources/PromiseKit.h
@@ -1,7 +1,12 @@
-#import "fwd.h"
-#import "AnyPromise.h"
-
 #import <Foundation/NSObjCRuntime.h>  // `FOUNDATION_EXPORT`
 
 FOUNDATION_EXPORT double PromiseKitVersionNumber;
 FOUNDATION_EXPORT const unsigned char PromiseKitVersionString[];
+
+// In this header, you should import all the public headers of your framework using statements like #import <PromiseKit/PublicHeader.h>
+
+#import <PromiseKit/AnyPromise.h>
+#import <PromiseKit/CLGeocoder+AnyPromise.h>
+#import <PromiseKit/CLLocationManager+AnyPromise.h>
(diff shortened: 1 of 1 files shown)


mxcl:
I cannot merge this because it imports headers that may not be in the final framework. Specifically the core location headers are a subspec that only some consumers specify.

dfed:
Got it! Welp. I'm shooting from the hip here, but maybe the right approach is to have each subspec add a compiler flag that we can use to conditionally import those files in the umbrella? I'd hate for subspecs to prevent us from having a proper umbrella header.

mxcl:
Yeah I'm pretty sure we can just let CocoaPods generate the header now.

If you change your PR to do that `.github/PromiseKit.podspec` is the file, and it passes CI, we should be good. (well that isn't a full lib lint, but if I can't push the pod then we'll know it's no good).

dfed:
Dumping the umbrella header might work. I know CocoaPods generates its own, and Buck will generate one if none exists. My issue right now is that Buck isn't generating one because we already have one. That said, removing the umbrella header will cause the build to break for anyone consuming PromiseKit with submodules.

That said, I'm curious what @dnkoutso thinks re the right approach for subspecs + umbrella headers, so, pinging him here :)

mxcl:
This will be unnecessary in v7 since I am removing all subspecs and making them their own pods.

v7 will however by Swift 5 only, so you may still want to apply the fix to v6.

dfed:
Seems reasonable. I'm happy to wait for v7 on this one. Will close this out.

PR #1012, opened 2019-02-13
Don't assume DispatchQueue.main === DispatchQueue.main

GarthSnyder:
Fixes test suite on Linux. The Travis build still does not pass the Deploy stage, but that looks like a different problem.


mxcl:
I squashed this into the bigger commit and rebased.

PR #1017, opened 2019-02-24
Dispatcher implementations

GarthSnyder:
This is not for direct integration yet, just review. 

I groomed the `Dispatcher` objects I ended up writing for another project and put them into a release-ready state. Take a look and see if you'd be interested in having these somewhere in the PromiseKit universe.

They're simple but probably too peripheral to put in the core; however, they might make a reasonable add-on package. I suspect that these few applications would cover 90% of the things people might be interested in using `Dispatcher` objects for in the real world.

Contents:

* `RateLimitedDispatcher` implements general dispatch rate limits with a token bucket (approximate) strategy.
* `StrictRateLimitedDispatcher` is an exact and optimal sliding window rate limiter, but requires O(n) space (n = # of events/interval)
* `ConcurrencyLimitedDispatcher` allows only *n* asynchronous closures to run at once.
* `CoreDataDispatcher` lets you dispatch onto threads associated with `NSManagedObjectContext`s.

`CoreDataDispatcher` is trivial. `ConcurrencyLimitedDispatcher` is just a few lines of code as well, although the exact way to implement it might not be obvious to those without some GCD knowledge.

The other two are a bit tricky to implement correctly and efficiently, so that's where I would locate the main value of releasing this stuff.


Sources/CoreDataDispatcher.swift
```diff
@@ -0,0 +1,31 @@
+#if !os(Linux)
```
mxcl: We can instead add this to the exclude list in Package.swift for this target.

Sources/ConcurrencyLimitedDispatcher.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+
+/// A PromiseKit Dispatcher that allows no more than X simultaneous
+/// executions at once.
+
+class ConcurrencyLimitedDispatcher: Dispatcher {
```
mxcl: Nice, we can remove the `when()` that does this manually. I figure. Haven't analyzed them both.

mxcl:
This is really great, and truly shows the value of your changes, my sincere thanks and respect!

PR #1024, opened 2019-03-21
v7 catch(only:) and recover(only:)

nathanhosselton:
This is #960 cherry picked onto the current `v7` branch.


nathanhosselton:
Whoops, I ran the tests on the master branch without paying attention. Obvs gotta fix some stuff up before it'll work on v7. Will do tomorrow probs.

mxcl:
The pretest failed, you have to run:

    swift test --generate-linuxmain

nathanhosselton:
Hmm, currently getting a `compiler is unable to type-check this expression in reasonable time` error on line 123 in the `testMapValues()` function of DispatcherTests.swift. Even with a clean clone of v7 (none of my commits). Unsure why, since I assume you are not. Using Xcode 10.2 beta 4.

Can attempt to fix but figured I'd check in first in case I'm simply missing something.

dougzilla32:
That’s mine I’ll take a look tonight.

dougzilla32:
I've created pull request #1027 to fix this problem.  My bad!

> Hmm, currently getting a `compiler is unable to type-check this expression in reasonable time` error on line 123 in the `testMapValues()` function of DispatcherTests.swift. Even with a clean clone of v7 (none of my commits). Unsure why, since I assume you are not. Using Xcode 10.2 beta 4.
> 
> Can attempt to fix but figured I'd check in first in case I'm simply missing something.

nathanhosselton:
I think this is good to go. I haven't studied v7 in depth but I believe I've made the appropriate changes in https://github.com/mxcl/PromiseKit/pull/1024/commits/3e58050062efe994fe0a7947ba33bee44d9577d9. lmk if not.

Force pushed with a replay of the commits because I accidentally left off the first commit when I opened this PR.

GarthSnyder:
The new variants of `catch` and `recover` need wrappers that accept a `DispatchQueue` instead of a `Dispatcher` as well. This is just to preserve the ability to write forms like `catch(only: FooError, on: .main)`. (This is new as of v7, so there's no reason you would have been aware of this.)

I'll submit a PR for these. I'm just mentioning it here in case the signatures for these methods change in the future.

mxcl:
As reviewer: my bad.

dougzilla32:
They also need cancellable varients.  I'm working on this now!

GarthSnyder:
Doug, I think we can pull out all the `catch` wrappers into a separate protocol and then just conform `Promise`, `CatchablePromise`, `PMKCascadingFinalizer`, and `CancellableCascadingFinalizer` (?) to it to avoid having to duplicate the wrappers. A start is in my catchonly-wrappers branch. The gist is:

```swift
public protocol CatchWrappers {
    associatedtype Finalizer
    associatedtype CascadingFinalizer
    // Base methods
    func `catch`(on: Dispatcher, policy: CatchPolicy, _ body: @escaping(Error) -> Void) -> Finalizer
    func `catch`<E: Swift.Error>(_ only: E, on: Dispatcher, _ body: @escaping() -> Void) -> CascadingFinalizer where E: Equatable
    func `catch`<E: Swift.Error>(_ only: E.Type, on: Dispatcher, policy: CatchPolicy, _ body: @escaping(E) -> Void) -> CascadingFinalizer
}

public extension CatchWrappers {
    // Wrapper implementations
} 

extension Promise: CatchWrappers {}  // Etc.
```

Probably something like this can be done for `recover`, too. I'm going to stop where I am for a while and await the cancellation updates. Don't worry about the wrappers - I can finish up that part when you're done.

dougzilla32:
Ah!  Brilliant!  That'll save a lot of boilerplate!!  And save us time. 👍 

I've actually just finished doing the wrappers so as you say I'll do the pull request and you can finish it up.  Thanks!!

GarthSnyder:
Nathan, one of the `recover` variants on `Thenable` doesn't accept a throwing closure. Is this intentional?

```swift
func recover<U: Thenable, E: Swift.Error>(_ only: E, on: Dispatcher = conf.D.map, _ body: @escaping() -> U) -> Promise<T> where U.T == T, E: Equatable
```

On another note, I initially put `error in` in my closure for this variant since that's what I was used to for the normal `recover`. I never saw such a series of misleading compiler errors! (E.g., it complained that the first argument should be labeled with `on:`).

I can see why there's no argument, but I wonder if it mightn't be better to add one just for consistency. The compiler seems to give more accurate error messages for cases where a closure ignores a supplied argument than the converse situation.

The same logic would apply to `catch`, of course.

mxcl:
Yeah it should allow throwing, and why is it `_ only:` this would indeed make it pretty ambiguous.

nathanhosselton:
Sorry for only just now responding to this.

IIRC, my logic for not allowing the `E` variants of recover (as opposed to the `E.Type` variants) to throw in their closures was due to knowing in advance which error occurred by virtue of the function. So, if you know the error, and you're recovering from it, there should be no reason to throw. Perhaps this was shortsighted of me though and there are cases where throwing would still be desired.

The parameter name being omitted `_ only:` was purely for succinctness. I agree with adding the name back if it would reduce compiler error ambiguity.

Indeed I omitted the error parameter of the closure for `E` variants of catch/recover because it felt redundant. But it also reduced ambiguity for the compiler (when they both had the error parameters, the compiler could not disambiguate them). Maybe this won't be an issue with the `only:` parameter name returned, though.

@mxcl @GarthSnyder If you'd like me to make any/all of these changes lmk. Happy to fix it up and submit the PR.

mxcl:
> IIRC, my logic for not allowing the E variants of recover (as opposed to the E.Type variants) to throw in their closures was due to knowing in advance which error occurred by virtue of the function. So, if you know the error, and you're recovering from it, there should be no reason to throw. Perhaps this was shortsighted of me though and there are cases where throwing would still be desired.

Smart, but well, imagine an HTTPError you may want to regrow for 403s, but not for any others.

> @mxcl @GarthSnyder If you'd like me to make any/all of these changes lmk. Happy to fix it up and submit the PR.

I would like them if @GarthSnyder agrees.

GarthSnyder:
I have used `recover` to transform one error into another, so I do think it's useful to be able to rethrow. I think my specific case was to intercept Core Data errors and stringify the details before reraising. The original errors contain live managed object references, so you can't let them propagate out to other threads - the original error wouldn't be legal to inspect there.

I added back the throwing in my clone already, just because I figured we'd probably do it and I wanted to include it in some test chains. (And as it happens, it's actually useful in that context since it lets you test multiple variants of `catch` in one chain.)

I don't have strong feeling about `_` vs. `only:` vs. `catchOnly`. That said, the current version (no explicit label required) reads pretty clearly already because the first argument is always going to be something type-explicit like `PMKError.badInput`.

Even though it was me that mentioned the surprise of not getting an `error` argument when filtering on a specific error, I don't think I feel strongly about that either. There is certainly something peculiar about receiving an argument whose value is already known; it sort of suggests that you need to check the value, which isn't true.

So Max, maybe you could make a final ruling on these two issues and I'll just fold them into my clone. Nathan can inspect the diffs at the next PR.

mxcl:
If it’s ambiguous then we should certainly add a named parameter, these ambiguity errors are the worst thing about PromiseKit.

GarthSnyder:
OK. From Nathan's comments above, it sounds like

- Including an `error` parameter for error-specific `recover` is a known ambiguity risk.
- The presence or absence of an `only:` label hasn't (yet) been seen to create a risk.

So it sounds like Nathan has already worked through this from the ambiguity perspective and that we should keep things as they are now.

nathanhosselton:
Opened #1038 for continued discussion/review.

PR #1028, opened 2019-03-25
'Cancel' for PromiseKit: add cancellable variants for #1024, & doc

dougzilla32:
* Add cancellable variants and cancellable dispatcher stubs for all new methods introduced by #1024
* Update Troubleshooting doc with solution for compilation timeouts, encountered primarily with cancellable promises


dougzilla32:
This isn't quite ready yet. While trying to write some tests to improve code coverage I discovered an ambiguous reference issue, still working on it.

dougzilla32:
This is in better shape now, although code coverage isn't quite where it should be.  I'll add some more tests when I get a chance.  The ambiguous reference issue was simply a case of specifying an incorrect return type for the closure.

This is ready to be merged other than adding a few more tests.

dougzilla32:
Yay, I got the code coverage check working locally so I don't need to rely on travis to generate the report.  I improved code coverage some more, although numbers are still slightly off.

If the numbers are good enough then we can merge; Garth is waiting on this one to do his dispatcher updates for #1024.

mxcl:
I thought I had configured codecov to allow up to a 0.3% drop. Apparently not.

GarthSnyder:
What became of this merge commit? I don't see it in the current `v7` history. 👀

I'm not seeing #1031, either. My clone is at `c33f`, which matches the commit shown here here for `v7`.

mxcl:
ugh, force pushed last night without pulling first. Will fix.

mxcl:
K fixed this locally, sorry for the delay, will push once v6 is passing after I fixed a variety of Swift 5 issues in the extensions.

mxcl:
FYI I rebased it all onto master preserving merges.

mxcl:
Pushed.

PR #1038, opened 2019-04-02
Improve usage of `catch(only:)` and `recover(only:)`

nathanhosselton:
Based on feedback in #1024, this PR attempts to reduce compiler ambiguity from these functions and allow for throwing from the closures of the `only: E` variants.

### Discussion
The final commit (bc81741) makes a change that, prior to this PR, would have created additional ambiguity. Requiring the `only:` parameter name (c81b91d) appears to have alleviated this, but including the error parameter in the closure comes with a tradeoff: while it stays consistent with other catch/recover functions, the parameter is likely to be unused or explicitly ignored in practice.

Although this change no longer has an effect on the ambiguity of the functions, forgetting to include the parameter (or if this commit isn't merged, forgetting to omit it) results in confusing compiler errors (e.g. "extra argument \`only:\` in call"). Due to this, it might make sense to include this change, if only because the case of forgetting to omit the parameter is probably more likely as there's precedent for its existence.

But if we don't want it I'll drop it and repush.


mxcl:
Being merged into #1045

GarthSnyder:
This was merged as part of #1045 - closing. Error-specific `catch` and `recover` seem to be working well. Thanks for the patch!

PR #1042, opened 2019-04-04
Changed module name

andredp:


andredp:
Sorry... this was a change I made to be able to migrate from PromiseKit4 to PK6 in our App (yep, we were still using PK4) and forgot to change the target branch to my local. 🙂

PR #1045, opened 2019-04-08
Wrapper revamp

GarthSnyder:
OK, now that the repo is straightened out, here's that wrapper reorganization that @dougzilla32 and I were talking about a while ago. The idea was to use protocols to collect wrapper signatures that were common to both promises and cancellable promises so that they didn't need to be duplicated. I also broke the wrappers out into a couple of separate files just to make them tidier.

Even though the wrapper protocols use `associatedtype`s to paper over the differences between cancellable and noncancellable promises, Xcode appears to see through these and show the normal types in popup help and documentation. If you can find cases where that's not so, I think that would be enough to scrap the whole idea, at least if it can't be worked around.

Unfortunately, there are still quite a few wrappers that can't be made common because `associatedtype`s cannot (yet) be generic. So anything that returns a `Promise<T>` or `CancellablePromise<T>` based on the `T` returned by a closure is unrepresentable in protocol land. 

I had earlier added @nathanhosselton's universal throwing changes to my clone, so I pulled Nathan's fork and disentangled us. This PR includes and subsumes #1038. I also replicated Nathan's changes into the cancellable promise and wrapper domains. I think it's all pretty complete, but Doug, you should probably review those changes to be sure I didn't screw it up.

A few other bits are in here:

* Added `cauterize()` for cascading finalizers

* Added a complete set of wrapper API tests. These aren't tests to exercise PromiseKit; they're just to make sure that all the right wrappers are defined and visible. There's a lot of duplication of existing forms, especially given the exhaustive tests that Doug added, but I figured the more independent test suites the better.

* Dispatcher type implementations (`RateLimitedDispatcher`, etc.) are now `final`. They were previously `public class`, which is just silly; they should either be `final` or `open`.

* Some copy editing for inline documentation - a lot of wrappers were missing mention of `flags`.

* I moved the files for cancellable promise support into a subdir.


dougzilla32:
I looked over your "wrapper-revamp" branch and it looks great to me! Thanks for cleaning this up!  Should be easier to maintain this way.

I see I forgot to rename CancellableTask to Cancellable -- good catch!

GarthSnyder:
Thanks, Doug - much appreciated!

Issue #1049, opened 2019-04-16
Some tests seem to depend on dispatching that PromiseKit should perhaps not be doing


GarthSnyder:
This is just a quick question for @dougzilla32...

Doug, there are multiple cancellation tests that seem to depend on calling `cancel()` at the end of a promise chain. Here's a typical one:

```swift
func testThenOffRejected() {
    let ex = expectation(description: "")
    CancellablePromise<Int>(error: PMKError.badInput).then { x -> Promise<Int> in
        XCTFail()
        return .value(x)
    }.catch(policy: .allErrors) {
        $0.isCancelled ? ex.fulfill() : XCTFail()
    }.cancel()
    wait(for: [ex], timeout: 1)
}
```

Since the promise starts out resolved, it seems to me that the `then` and the `catch` should both be resolved at creation time as well. The body of the `catch` has to be dispatched, but the argument to that closure should be the original `badInput` error, because the `catch` closure should be dispatched before `cancel` even runs. Or at least, that's my admittedly dim understanding.

I think the reason it _doesn't_ work that way at the moment is because of some wrapper code in one of the `CancellablePromise` `init` implementations (a version that wraps an existing promise):

```swift
if promise == nil {
    // Wrapper promise
    promise = Promise { seal in
        reject = seal.reject
        bridge.done(on: CurrentThreadDispatcher()) {
            seal.fulfill($0)
        }.catch(policy: .allErrors) {
            seal.reject($0)
        }
    }
}
``` 

This inner `catch` is dispatched, so that creates a stall in the pipeline that gives the `cancel` operation an opportunity to swoop in.

Since this is something that several tests seem to be explicitly checking for, I'm wondering if it's part of the intended API that `cancel` is in some sense "faster" than normal cascading. Would it be OK to add `on: CurrentThreadDispatcher()` to the wrapper `catch` and to adjust the tests so that they don't assume `cancel` will be dominant, or is that likely to break something important in the cancellation system?

dougzilla32:
Ah!  I did notice this while working on the tests and didn't know why I was able to cancel a promise that starts out as resolved.  But I forgot to go back later and figure out what is going on.

Definitely this is a bug and your fix looks great!!  I'm working on a pull request to incorporate your fix and update the tests.

Thank you for catching this!!

dougzilla32:
Created pull request #1050 to address this.  Garth, lmk if this looks good to you and thanks again for the feedback!

GarthSnyder:
[...]

PR #1050, opened 2019-04-20
CancellablePromise bridge initializer should run 'catch' on CurrentThreadDispatcher()

dougzilla32:
Fix for #1049: "Some tests seem to depend on dispatching that PromiseKit should perhaps not be doing"

Yup, CancellablePromise is definitely doing some dispatching that it shouldn't be doing.  Fixed exactly as Garth suggests, by having the CancellablePromise bridge initializer run 'catch' on CurrentThreadDispatcher().

Also updating 6 affected tests for this change.


GarthSnyder:
Looks awesome, Doug - thanks for taking a look at this!

GarthSnyder:
There's a 7.0.0-alpha3 tag at the previous head which I didn't move or supersede.

PR #1053, opened 2019-05-01
Replacing Result enum in favor of the one available in Swift 5

heyzooi:


mxcl:
We cannot add support to v6 without breaking API compatibility. v7 is in alpha and already uses Swift 5’s result.

PR #1058, opened 2019-05-17
‘firstly’ method extended 

djmixroman:


mxcl:
This has been requested in the past, and we’ve rejected it because it makes `firstly` behave differently in a “zalgo” sense: https://blog.izs.me/2013/08/designing-apis-for-asynchrony

The default `firstly` executes immediately, while these additions would not.

And, using `DispatchQueue.foo.async(.promise)` is not too arduous, though not as nice for sure.

Thank you though. Sorry you wasted your time.

PR #1065, opened 2019-05-31
Update Appendix.md

paulthorsteinson:
As "PromiseKit has perhaps capriciously switched between the names always and ensure", this page just needs an update to refer to ensure instead of always :)


mxcl:
Thanks!

Issue #1068, opened 2019-06-07
compactMapValues/flatMapValues/mapValues for Guarantee


Igor-Palaguta:
Probably Sequence functions can be added for Guarantees also?

Can prepare PR if needed

mxcl:
Yeah PR welcome, this is an omission for sure.

RomanPodymov:
I think the issue should be closed.

PR #1074, opened 2019-06-17
Add guarantee sequence utils

Igor-Palaguta:
Fix for https://github.com/mxcl/PromiseKit/issues/1068


mxcl:
Great, thanks. You may want to add these to `v7` also, or they may vanish when we release it.

GarthSnyder:
I'll port these to v7 - probably needs some extra wrappers.

PR #1092, opened 2019-08-24
when(resolved:) tuple overloads

Skoti:
This PR introduces tuple overloads for `when(resolved:)` (which make use of `Thenable` protocol) and additional tuple overloads for `when(fulfilled:)` up to 9 arguments.
This PR also introduces an associated type `R` representing the result type in `Thenable` protocol which improves a return type in `when(resolved:)` when using Guarantees, so this is now allowed:
```
let p1 = Promise.value(true)
let p2 = Promise.value(2)
let g1 = Guarantee.value("abc")

when(resolved: p1, p2, g1).done { r1, r2, r3 in
    // r1 is Result<Bool>, r2 is Result<Int>, r3 is String
}
```

In the future it might be worth considering changing the `pipe` declaration to use `R` instead of `Result<T>`, this would however impact all the other operators so I didn't want to introduce such a big change in this PR.


mxcl:
Seems cool. Can we make tests pass?

Skoti:
Done :)

Skoti:
still interested in merging? :)

IgnusG:
This would be great - is there a timeline on this or is PMK7 incompatible/already implementing this?

Tests/CorePromise/WhenResolvedTests.swift
```diff
@@ -5,11 +5,12 @@ import PromiseKit
 import XCTest
 
 class JoinTests: XCTestCase {
+    
     func testImmediates() {
         let successPromise = Promise()
 
         var joinFinished = false
-        when(resolved: successPromise).done(on: nil) { _ in joinFinished = true }
+        when(resolved: [successPromise]).done(on: nil) { joinFinished = true }
```
mxcl: So, does this mean we have broken API? We can't break API.

Tests/CorePromise/WhenTests.swift
```diff
@@ -5,22 +5,24 @@ import XCTest
...
+        let e1 = expectation(description: "")
         when(fulfilled: promises).done { _ in
             e1.fulfill()
         }.silenceWarning()
 
         let e2 = expectation(description: "")
         when(resolved: promises).done { _ in
             e2.fulfill()
-        }.silenceWarning()
+        }
```
mxcl: Does this mean the `when` has `@discardableResult` when it should not?

Tests/CorePromise/WhenResolvedTests.swift
Skoti: This is because of "ambiguous use of 'when'" errors. I have commented out the responsible functions for now:
https://github.com/Skoti/PromiseKit/blob/d13d220396cf88d649d02ef82549eeb85cc95816/Sources/when.swift#L86-L89
and here:
https://github.com/Skoti/PromiseKit/blob/d13d220396cf88d649d02ef82549eeb85cc95816/Sources/when.swift#L302-L305
Left them commented out for a reminder.

It seems like adding this `R` associated type representing result in `Thenable` protocol confuses the compiler for some reason? 🤷‍♂ As before that it was not a problem.

Anyway, I think tuple overloads with this `R` associated type give us much more flexibility of using Promises and Guarantees as well as mixing them and a single use case of a convenience `Void` variadic function is not taking these advantages down. Plus it is a really easy fix for such a breaking change - you just add `[` and `]` around the arguments...

Tests/CorePromise/WhenTests.swift
Skoti: It is not about `when`, but about `done`. 
`when(fulfilled:)` returns Promise and `done` for Thenable is declared **without** `@discardableResult`, however, `when(resolved:)` returns Guarantee and `done` for Guarantee is declared **with** `@discardableResult`.
That is why it does not need `.silenceWarning()`. It has **not** been changed in this PR.

Thenable:
https://github.com/mxcl/PromiseKit/blob/f8a593a320401045979e705640e654f6c2e0bba3/Sources/Thenable.swift#L167-L168

Guarantee:
https://github.com/mxcl/PromiseKit/blob/f8a593a320401045979e705640e654f6c2e0bba3/Sources/Guarantee.swift#L78-L80

Tests/CorePromise/WhenResolvedTests.swift
mxcl: I agree this is better, but breaking API is not allowed, API breakage can only happen for major version bumps, so this may have to be a v7 PR if we cannot solve it.

Tests/CorePromise/WhenTests.swift
mxcl: Oh, I'm a dumbo, thanks.

Tests/CorePromise/WhenResolvedTests.swift
Skoti: Ok! Just let me know when you are ready with the v7 changes. Then I will rebase this PR and you can review it and possibly merge before v7 release 😄

mxcl: v7 is ready for submissions.

Skoti: hmm but the v7 branch was updated 6-7 months ago, is this is the one I should use?

mxcl:
If we can base this onto v7 we can get it in.

Skoti:
> If we can base this onto v7 we can get it in.

Ok great! Give me some time :)

Skoti:
I've rebased this PR and added tuple arity functions for cancellable thenables up to 9 arguments.
I've also improved `when(resolved:)` for array and `when(fulfilled:) with generator to work for any `CancellableThenable` and not only `CancellablePromise`.
I've added `when(resolved:)` with generator.
I've renamed `cancellableWhen` functions to just `when` to match the naming convention.
In order to keep the tests more readable and clean I've added a shared `TestError` and `TestCancelError` types.

mxcl:
Merge conflicts.

Skoti:
The PR was still set to merge to `master`. I've just changed the base branch to `v7`.

Tests/Cancel/ResolverTests.swift
```diff
@@ -46,39 +47,39 @@ class WrapTests: XCTestCase {
         }
         
         let value: Int?
-        let error: Swift.Error?
```
mxcl, line 49: These changes are unnecessary.

Tests/Core/CancellableErrorTests.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 import PromiseKit
+import TestsShared
```
mxcl, line 3: What is this new module for?

Tests/PMKCoreLocation/CLGeocoderTests.swift
```diff
@@ -17,7 +17,7 @@ class CLGeocoderTests: XCTestCase {
         }
 
         let ex = expectation(description: "")
-        MockGeocoder().reverseGeocode(location: CLLocation()).done { x in
+        _ = MockGeocoder().reverseGeocode(location: CLLocation()).done { x in
```
mxcl, line 20: Let’s not do unnecessary things in PRs thanks. Tight PRs without other (even well intentioned) changes prevent review errors.

mxcl:
CI did not run (GitHub Actions bug perhaps) we need that to verify that the changes do not cause ambiguity. Even then I am concerned about the change of `cancellableWhen` to `when` being ambiguous to the compiler.

I'd like to see this PR without all the infrastructure changes since it bloats it substantially.

Tests/Core/CancellableErrorTests.swift
Skoti: The new module is to have a common set of files for tests targets, i.e. `TestError` is there so you do not need to declare dummy errors everywhere as it makes things confusing and tests harder to read.

Tests/PMKCoreLocation/CLGeocoderTests.swift
Skoti: this is just to get rid of the warning, but ok I can revert it

Skoti:
The ambiguity usages were due to the variadic `when` functions.

>I agree this is better, but breaking API is not allowed, API breakage can only happen for major version bumps, so this may have to be a v7 PR if we cannot solve it.

But we agreed that it is better to have tuple arity functions, as variadic is easy to fix as you just add [ and ] around the arguments. From the calling point of view it is almost no difference. Not sure what is more convenient with variadic arguments that an array couldn't handle.

By infrastructure you mean adding a common module?

Skoti:
@mxcl I got rid of the infrastructure changes.

Sources/PromiseKit/Cancellation/CancellableThenable.swift
```diff
@@ -296,48 +296,55 @@ public extension CancellableThenable {
...
-     - Returns: `true` if the cancellable promise has not yet resolved.
+     - `true` if the cancellable promise has not yet resolved.
      */
     var isPending: Bool {
         return thenable.isPending
     }
 
     /**
-     - Returns: `true` if the cancellable promise has resolved.
+     - `true` if the cancellable promise has resolved.
```
mxcl, line 306: Why are we removing the documentation markers? This will cause our generated documentation to render incorrectly.

Skoti: Actually, in Xcode it shows incorrectly when using `Returns` keyword. This is a computed property and it does not use this keyword.
With `Returns:`
with
Without `Returns:`
without
Now I think we could even remove the dash "-":
dash
What do you think? Do you use any other tool for generating documentation? I think Xcode quick-help documentation should be prioritized.

mxcl: Jazzy renders it:

https://mxcl.dev/PromiseKit/reference/v7/Protocols/CancellableThenable.html#/s:10PromiseKit19CancellableThenablePAAE5value1U_1TQZSgvp

https://github.com/mxcl/PromiseKit/blob/master/Sources/PromiseKit/Cancellation/CancellableThenable.swift#L333-L338

If we need to fix documentation it needs to be a separate PR.

While I appreciate this is more work for contribution, my main priority with this project is maintaining quality, not encouraging contribution. Your contributions are sincerely appreciated of course, but this project has tens of thousands of users, so my obligation is to them.

Skoti: But it will still work with jazzy, it's a win-win change.
Result is without dash, value is with a dash:
jazzy

mxcl: According to docs I could find `- Returns:` is correct, if Xcode doesn’t render it, it's an Xcode bug.

Please keep PRs focused on their task, tidying things up and fixing other things should be separate PRs to prevent the PR being 2 months of discussion on unrelated changes before we can actually review the code changes.

Case in point, this discussion here.

Skoti: As you wish.

Skoti:
Thanks for revisiting this PR.

Honestly, I can't see anything similar. Not much has changed since then.
The `v8` is just a continuation of `v6` with little changes, and the `v7` was never released.

I think this PR brings a lot of benefits, but because the `v7` was skipped it would need to be `v9` now.

Anyway, in the modern Swift with `Combine` and `async/await` many people have switched to using them, and I don't think there is yet a place for `PromiseKit`, I also haven't used PromiseKit for a long time.

Please do what you think is best.

PR #1096, opened 2019-09-24
Promise.value and Guarantee.value when T == Void

RomanPodymov:
Hello.
Thank you for PromiseKit.
I see that `Resolver` has `func fulfill_()` for `T == Void`. I would propose to add similar functions for `Promise` and `Guarantee`.
Also, declared `func Promise.value()` as `static` because `Promise` is `final`.


mxcl:
Thanks, this is a good addition.

Can we get a test that verifies that this is not ambiguous for any version of Swift we support?

ie:

```swift
func test() {
    // tests no amiguity
    let a = Promise<Void>.value
    let b = Promise<Void>.value(Void())
}
```

Thanks.

RomanPodymov:
> Thanks, this is a good addition.
> 
> Can we get a test that verifies that this is not ambiguous for any version of Swift we support?
> 
> ie:
> 
> ```swift
> func test() {
>     // tests no amiguity
>     let a = Promise<Void>.value
>     let b = Promise<Void>.value(Void())
> }
> ```
> 
> Thanks.

Hello @mxcl 
Done.

mxcl:
Marvelous thank you for your contribution!

PR #1101, opened 2019-10-05
Fixed possible "Ambiguous use of 'when(fulfilled:)'"

RomanPodymov:
Hello.
Thank you for PromiseKit.
Found this test 
```
func testAllSealedRejectedFirstOneRejects() {
    enum Error: Swift.Error {
        case test1
        case test2
        case test3
    }

    let ex = expectation(description: "")
    let p1 = Promise<Void>(error: Error.test1)
    let p2 = Promise<Void>(error: Error.test2)
    let p3 = Promise<Void>(error: Error.test3)

    when(fulfilled: p1, p2, p3).catch { error in
        XCTAssertTrue(error as? Error == Error.test1)
        ex.fulfill()
    }

    waitForExpectations(timeout: 1)
}
```
Which function is calling on `when(fulfilled: p1, p2, p3)`? `func when<U: Thenable>(fulfilled thenables: U...) -> Promise<Void> where U.T == Void` or `func when<U: Thenable, V: Thenable, W: Thenable>(fulfilled pu: U, _ pv: V, _ pw: W) -> Promise<(U.T, V.T, W.T)>`? To avoid possible **"Ambiguous use of 'when(fulfilled:)'"** I offer to use another method signature for some cases. Also see [this issue](https://bugs.swift.org/browse/SR-11572) for more details.

### DIFF
--- a/Sources/when.swift
+++ b/Sources/when.swift
@@ -67,36 +67,36 @@ private func _when<U: Thenable>(_ thenables: [U]) -> Promise<Void> {
  - SeeAlso: `when(resolved:)`
 */
 public func when<U: Thenable>(fulfilled thenables: [U]) -> Promise<[U.T]> {
-    return _when(thenables).map(on: nil) { thenables.map{ $0.value! } }
+    return _when(thenables).map(on: nil) { thenables.map { $0.value! } }
 }
 
 /// Wait for all promises in a set to fulfill.
-public func when<U: Thenable>(fulfilled promises: U...) -> Promise<Void> where U.T == Void {
-    return _when(promises)
+public func when<U: Thenable>(fulfilled thenables: U...) -> Promise<Void> where U.T == Void {
+    return _when(thenables)
 }
 
 /// Wait for all promises in a set to fulfill.
-public func when<U: Thenable>(fulfilled promises: [U]) -> Promise<Void> where U.T == Void {
-    return _when(promises)
+public func when<U: Thenable>(fulfilled thenables: [U]) -> Promise<Void> where U.T == Void {
+    return _when(thenables)
 }
 
 /// Wait for all promises in a set to fulfill.
-public func when<U: Thenable, V: Thenable>(fulfilled pu: U, _ pv: V) -> Promise<(U.T, V.T)> {
+public func when<U: Thenable, V: Thenable>(thenables pu: U, _ pv: V) -> Promise<(U.T, V.T)> {
     return _when([pu.asVoid(), pv.asVoid()]).map(on: nil) { (pu.value!, pv.value!) }
 }
 
 /// Wait for all promises in a set to fulfill.
-public func when<U: Thenable, V: Thenable, W: Thenable>(fulfilled pu: U, _ pv: V, _ pw: W) -> Promise<(U.T, V.T, W.T)> {
+public func when<U: Thenable, V: Thenable, W: Thenable>(thenables pu: U, _ pv: V, _ pw: W) -> Promise<(U.T, V.T, W.T)> {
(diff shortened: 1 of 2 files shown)


mxcl:
Isn’t this API breaking? We cannot remove some instances `when(fulfilled:)`, unless we hate our users.

RomanPodymov:
> Isn’t this API breaking? We cannot remove some instances `when(fulfilled:)`, unless we hate our users.

I planned to add some new versions of `when` for `Guarantee`s:

```
private func _when<U>(_ guarantees: [Guarantee<U>]) -> Guarantee<Void> {
    var countdown = guarantees.count
    guard countdown > 0 else {
        return .value(Void())
    }

    let rp = Guarantee<Void>(.pending)

#if PMKDisableProgress || os(Linux)
    var progress: (completedUnitCount: Int, totalUnitCount: Int) = (0, 0)
#else
    let progress = Progress(totalUnitCount: Int64(guarantees.count))
    progress.isCancellable = false
    progress.isPausable = false
#endif

    let barrier = DispatchQueue(label: "org.promisekit.barrier.when", attributes: .concurrent)

    for guarantee in guarantees {
        guarantee.pipe { (_: U) in
            barrier.sync(flags: .barrier) {
                guard rp.isPending else { return }
                progress.completedUnitCount += 1
                countdown -= 1
                if countdown == 0 {
                    rp.box.seal(())
                }
            }
        }
    }

    return rp
}

public func when<U, V>(guarantees pu: Guarantee<U>, _ pv: Guarantee<V>) -> Guarantee<(U, V)> {
    return _when([pu.asVoid(), pv.asVoid()]).map(on: nil) { (pu.value!, pv.value!) }
}

public func when<U, V, W>(guarantees pu: Guarantee<U>, _ pv: Guarantee<V>, _ pw: Guarantee<W>) -> Guarantee<(U, V, W)> {
    return _when([pu.asVoid(), pv.asVoid(), pw.asVoid()]).map(on: nil) { (pu.value!, pv.value!, pw.value!) }
}

public func when<U, V, W, X>(guarantees pu: Guarantee<U>, _ pv: Guarantee<V>, _ pw: Guarantee<W>, _ px: Guarantee<X>) -> Guarantee<(U, V, W, X)> {
    return _when([pu.asVoid(), pv.asVoid(), pw.asVoid(), px.asVoid()]).map(on: nil) { (pu.value!, pv.value!, pw.value!, px.value!) }
}

public func when<U, V, W, X, Y>(guarantees pu: Guarantee<U>, _ pv: Guarantee<V>, _ pw: Guarantee<W>, _ px: Guarantee<X>, _ py: Guarantee<Y>) -> Guarantee<(U, V, W, X, Y)> {
    return _when([pu.asVoid(), pv.asVoid(), pw.asVoid(), px.asVoid(), py.asVoid()]).map(on: nil) { (pu.value!, pv.value!, pw.value!, px.value!, py.value!) }
}
```

However I cannot compile this code because getting **Ambiguous use of 'when'** on `testAllSealedRejectedFirstOneRejects`. I think this is a [Swift compiler bug](https://bugs.swift.org/browse/SR-11572).
I know that my pull requests breaks API, but how should we solve the problem? I want to create another pull request with new `when` versions, but should I rename `when` to `whenGuarantees`?

mxcl:
Yeah, if we cannot avoid breaking API, then it has to go in v7. Otherwise all 100,000 PMK users may/will get breakage.

RomanPodymov:
> Yeah, if we cannot avoid breaking API, then it has to go in v7. Otherwise all 100,000 PMK users may/will get breakage.

OK, I will create another pull request to v7.

PR #1112, opened 2019-11-13
refactor(*): Change empty tuples to `Void`

emrcftci:
# Description

We have a typealias for empty tuples `Void`. 
This pull-request changes empty tuples `()` to `Void` in return types.

# Motivation and Context

Always used `Void` in return types in this project so I change to recover this in forgotten functions and closures.

PTAL @mxcl 🚀 📱 And thank you for your dedication to the open source. 🙏


mxcl:
Thanks! This is indeed per style guides.

PR #1113, opened 2019-11-15
Documentation: Clarify when a promise "starts" in the FAQ

jfahrenkrug:
I was confused when reading the FAQ about when promises "start". I'm aware that this PR will most likely be the basis for a discussion and might invoke polarizing opinions. Maybe my change is completely wrong ;-) Please correct me, but I do think we can and should clarify and simplify this FAQ item.

Provide a simpler explanation and examples about when promises "start".
Make it clear that the promise's body executes after the promise is created,
without the need to call `then` or `done` on it. Also clarify that async
tasks that a being kicked off from a promise's body behave the same way
regardless of whether PromiseKit is being used or not.

### DIFF
--- a/Documentation/FAQ.md
+++ b/Documentation/FAQ.md
@@ -258,7 +258,7 @@ You want `recover`.
 Often people are confused about when Promises “start”. Is it immediately? Is it
 later? Is it when you call `then`?
 
-The answer is: The promise **body** executes right after the promise is created, on the current thread.
+The answer is: The promise **body** executes during initialization of the promise, on the current thread.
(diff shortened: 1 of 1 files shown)


mxcl:
Nice, this is clearer IMO, however, I think it should say:

> The promise **body** executes during initialization of the promise

Rather than:

>The promise **body** executes right after the promise is created

Since it is *not* executed after.

jfahrenkrug:
@mxcl Thanks for the quick feedback, I'll fix it right away, thanks for clarifying!

mxcl:
Thank you.

PR #1117, opened 2019-12-19
Add module stability by enabling library evolution support

yo2bh:
Enabled the BUILD_LIBRARY_FOR_DISTRIBUTION flag introduced in Xcode 11 for supporting framework distribution for creating the XCFramework.


mxcl:
Seems to break Swift 4.1.52 compilation for some reason. See Travis.

https://travis-ci.org/mxcl/PromiseKit/jobs/627150231

RomanPodymov:
> Seems to break Swift 4.1.52 compilation for some reason. See Travis.
> 
> https://travis-ci.org/mxcl/PromiseKit/jobs/627150231

I think I had the same problem in https://github.com/mxcl/PromiseKit/pull/1121. Probably we should restart the checks.

mxcl:
I restarted it twice already, but I’ll try again.

mxcl:
Tried 4 times now, seems consistent.

mxcl:
Notably this is the only build that is Xcode 11 and uses xcodebuild to build it, so basically, seems like this flag fails.

yo2bh:
@mxcl Is there any way to resolve this issue which we're facing in Travis CI?

mxcl:
I am unfamiliar with the `BUILD_LIBRARY_FOR_DISTRIBUTION` so don’t know. Certainly not being able to build against a platform we support will prevent us merging this.

yo2bh:
@mxcl - This flag is used to enabling library evolution support.
i.e Swift Library Evolution makes it so that you can link an application against one version of a Swift library, then upgrade the library to a newer version without rebuilding the application.

and 

one more thing to create XCFramework, BUILD_LIBRARY_FOR_DISTRIBUTION flag must be set in a build setting.

We can set this flag in Podspec as well but I did not find the.podspec file in the repo.

@mxcl - Can you please help me out for this.

mxcl:
Yeah I read about this recently at the Swift blog.

I don’t really know what to say, if it won’t build against a platform we support I cannot merge it into v6.

RomanPodymov:
Look what I found at [Library Evolution in Swift](https://swift.org/blog/library-evolution):

> Library evolution support should only be used when a framework is going to be built and updated separately from its clients.

PromiseKit is an open source project, so it does not make sense to build it with `BUILD_LIBRARY_FOR_DISTRIBUTION = YES`.

yo2bh:
@RomanPodmov & @mxcl -

What should we do if we have to going build separately to the client ? Is there any way to do this ?

mxcl:
> What should we do if we have to going build separately to the client? Is there any way to do this?

Modify the xcodeproj with a script.

RomanPodymov:
Hello @yo2bh 
You can use https://github.com/CocoaPods/Xcodeproj and set `BUILD_LIBRARY_FOR_DISTRIBUTION = YES` for any project. But for PromiseKit let's keep `BUILD_LIBRARY_FOR_DISTRIBUTION = NO` like in other open source projects.

PR #1118, opened 2019-12-28
Map by keyPath

RomanPodymov:
Hello.
Thank you for PromiseKit.
In this pull request I added new versions of `map` for `Thenable` and `Guarantee`. Now you can use a key path instead of a closure if you need to map a value.


mxcl:
Odd the code coverage dropped.

PR #1121, opened 2020-01-08
More map by keyPath

RomanPodymov:
Hello.
Thank you for PromiseKit.
In this pull request I added more mapping functions that accepts `KeyPath` instead of a closure (started in #1118).


mxcl:
I have confirmed that Swift 5.2 adds the SE proposal that KeyPaths are compiler converted to functions with literal syntax, thus this addition is welcome, but should be guarded with `#if swift(<5.2)` to prevent collision.

RomanPodymov:
Hello @mxcl 
I see that the Travis CI build could not complete due to an error. How can I restart it?

Issue #1114, opened 2019-11-23
Request: Wait for one promise to fulfill while ignoring rejections


pimnijman:
I have multiple promises that can be pending for a long time, or get rejected early on. I'm looking for a way to find out which is the first to fulfill.

For example: One promise gets rejected, another fulfills and a third is still pending.
In this case I would like the new promise to fulfill.
When all promises get rejected I would like the new promise to reject with an array of errors.

I can't use `race`, since it will reject once my first promise will reject.
I can't use `when(fulfilled:)`, since it will reject once one of my promise rejects and will continue after the first promise fulfills.
I can't use `when(resolved:)`, since it will continue after the first promise resolves.

What I'm looking for might not exist (yet) but I would expect it to be called something like  `race(fulfilled:)`

mxcl:
`when(resolved:)` may work, but it still waits for all promises. `race` will work, hold onto the promises and check each one before deciding what to do.

icedice:
> `when(resolved:)` may work, but it still waits for all promises. `race` will work, hold onto the promises and check each one before deciding what to do.

Could you show an example on how to this? I'm using race for caching. I want to call cache and network at the same time and then return what ever finish first, however right now if cache is empty and the promise rejects I will not get the result from the network.

mxcl:
Implemented.

PR #1123, opened 2020-01-19
Added when(maxFulfilledCount:)

RomanPodymov:
Hello.
Thank you for PromiseKit.
In this pull request I added a new parameter to some `when` versions. With this parameter `when` behaves like `race` with the amount of winners.

### DIFF
--- a/Sources/when.swift
+++ b/Sources/when.swift
@@ -1,8 +1,8 @@
 import Foundation
 import Dispatch
 
-private func _when<U: Thenable>(_ thenables: [U]) -> Promise<Void> {
-    var countdown = thenables.count
+private func _when<U: Thenable>(_ thenables: [U], maxFulfilledCount: Int? = nil) -> Promise<Void> {
+    var countdown = min(maxFulfilledCount ?? thenables.count, thenables.count)
     guard countdown > 0 else {
         return .value(Void())
     }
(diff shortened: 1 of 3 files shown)


mxcl:
1. What is this for? Please provide concrete examples of where it is useful
2. We should not overload `when` with this, unless it makes sense to, but I don’t see how it makes sense idiomatically as yet

As a note, generally it is advisable to discuss adding features to a library that is almost 10 years old before investing time making a PR. This PR is high quality, but the reason PMK is high quality is because we are careful about additions.

RomanPodymov:
@mxcl I was thinking about the solution for #1114. As for me there should be a function like `when(resolved:maxFulfilledCount:)` and it should be called like `when(resolved: [fulfilledPromise, errorPromise, pendingPromise], maxFulfilledCount: 1)`. In this pull request I added a similar function that can be also useful. Could you please explain more properly how can we resolve #1114 without a new function?

mxcl:
I feel a variant of `race` is more what you want to provide, perhaps `race(fulfilled:)` which implies it allows failures and waits for some promise to actually `win`.

RomanPodymov:
Hello @mxcl 
I opened [a pull request](#1125) for `race(fulfilled:)`.

PR #1132, opened 2020-03-05
Add support for policy param in Promise<Void>.recover -> Guarantee

diesal11:


mxcl:
This is not logically possible, `Guarantee` cannot contain an error, so it cannot pass through cancelations.

diesal11:
Ah! I sincerely apologise, i thought i was adding support for catching `PMKError.cancelled` errors in here, but it already passes them in without this param.

mxcl:
No worries, I think the documentation is not clear, so that’s my bad.

Issue #1137, opened 2020-03-26
Getting warning Module 'PromiseKit' was not compiled with library evolution support


yo2bh:
I am getting below warning while compiling the application.

_"Module 'PromiseKit' was not compiled with library evolution support; using it means binary compatibility for 'MyFramework' can't be guaranteed"_

Following are the Promisekit details:
_Version_ - 6.8.4
_Installed_ - Carthage.

How to resolve this warning so that this will not reflect in warning?

martinpucik:
@yo2bh did you try to rebuild your Carthage dependencies with the latest Xcode?

yo2bh:
@martinpucik - I am using latest version of Xcode 11.3. Still facing the same issue.

bielikb:
@yo2bh You're seeing this warning since you're building framework with module stability turned on == you're setting `BUILD_LIBRARY_FOR_DISTRIBUTION=YES`  in your build settings/xcconfig.

This flag shouldn't be pushed to all `PromiseKit` users, though, since it's not required for schemes that do build apps.

To resolve this issue, I'd suggest trying to build/update the `PromiseKit` dependency with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` flag. If you'll hit any issues with turning the module stability on in the `PromiseKit`, open another issue that will reflect that specific issue.

_Note:_
I've written couple words about the xcframeworks & module stabillity here https://github.com/bielikb/xcframeworks, including [Testing & Troubleshooting](https://github.com/bielikb/xcframeworks#testing--troubleshooting) section

RomanPodymov:
Hello @yo2bh 
Can you reproduce this issue in Xcode 12?

mxcl:
Presumably no longer an issue.

PR #1143, opened 2020-04-30
Enable library evolution

pepone:
Hi,

This PR enables "Library Evolution" support

See https://swift.org/blog/abi-stability-and-more/

We (ZeroC)  are trying to enable this for [zeroc-ice/ice](https://github.com/zeroc-ice/ice) but we are getting errors because PromiseKit doesn't enable it.

```
error: module 'PromiseKit' was not compiled with library evolution support; using it means binary compatibility for 'Ice' can't be guaranteed
import PromiseKit
```

It will be nice if this can be merge and get Library Evolution enabled for the next PromiseKit release.

I think this will also solve #1137
Cheers,
José


martinpucik:
I don't think this should be enabled for PromiseKit, afaik `BUILD_LIBRARY_FOR_DISTRIBUTION` is not intended for libraries distributed via e.g. Cocoapods and embeded inside the app.
More about this in conversation from https://github.com/CocoaPods/CocoaPods/issues/9232#issuecomment-566713762

#1137 should be resolved by running `carthage bootstrap`, similar issue discussed here https://github.com/airbnb/lottie-ios/pull/1173#issuecomment-610511321

RomanPodymov:
Hello @pepone 
I think this pull request is a duplicate for #1117.

pepone:
Hi @RomanPodymov

Thanks I miss that, I closing this

PR #1144, opened 2020-04-30
Remove unused variable

pepone:
This is just a minor fix to remove an unused variable, I also enable the "Treat Warnings as Error" option for PromiseKit target.

### DIFF
--- a/PromiseKit.xcodeproj/project.pbxproj
+++ b/PromiseKit.xcodeproj/project.pbxproj
@@ -1071,7 +1071,7 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				PRODUCT_BUNDLE_PACKAGE_TYPE = FMWK;
 				SKIP_INSTALL = YES;
-				SWIFT_TREAT_WARNINGS_AS_ERRORS = NO;
+				SWIFT_TREAT_WARNINGS_AS_ERRORS = YES;
 				TARGETED_DEVICE_FAMILY = "1,2,3,4";
 			};
 			name = Debug;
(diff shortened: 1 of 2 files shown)


guillaumealgis:
I don't necessarily agree with the "Treat Warnings as Error" bit, but fixing the warning for the unused variable is great, thanks :)

mxcl:
Yeah, we will regret treat warnings as errors when swift introduces a new warning and Carthage users cannot compile our package, that has to go I'm afraid.

Sources/when.swift
```diff
@@ -154,14 +154,12 @@ public func when<It: IteratorProtocol>(fulfilled promiseIterator: It, concurrent
         }
         guard shouldDequeue else { return }
 
-        var index: Int!
```
RomanPodymov, line 157: I think it was already fixed in #1154.

PromiseKit.xcodeproj/project.pbxproj
```diff
@@ -1071,7 +1071,7 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				PRODUCT_BUNDLE_PACKAGE_TYPE = FMWK;
 				SKIP_INSTALL = YES;
-				SWIFT_TREAT_WARNINGS_AS_ERRORS = NO;
+				SWIFT_TREAT_WARNINGS_AS_ERRORS = YES;
```
RomanPodymov, line 1074: As you can see at https://github.com/mxcl/PromiseKit/pull/1144#issuecomment-637866172 you should keep `SWIFT_TREAT_WARNINGS_AS_ERRORS = NO;`.

RomanPodymov (commented):
I think this pull request should be closed.

PR #1160, opened 2020-08-27
Updated product to be dynamic

vpolsinelli:
Switched the products from Swift Package Manager to be `dynamic` rather than `static`. 

By default, Swift Package Manager sets this to `static` and can cause linking issues. The specific error I am trying to address is:
> error: Swift package product LIBRARY is linked as a static library by APP and EXTENSION. This will result in duplication of library code.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,7 @@ import PackageDescription
 
 let pkg = Package(name: "PromiseKit")
 pkg.products = [
-    .library(name: "PromiseKit", targets: ["PromiseKit"]),
+    .library(name: "PromiseKit", type: .dynamic, targets: ["PromiseKit"]),
 ]
 
 let pmk: Target = .target(name: "PromiseKit")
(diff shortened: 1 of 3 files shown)


RomanPodymov:
Hello @vpolsinelli 
Can you reproduce the issue in Xcode 11.7?

vpolsinelli:
Hey @RomanPodymov,

I am still getting this error in Xcode 11.7.

The dependency structure I have is as follows:
App
- PromiseKit
- Dependency A
    - PromiseKit
- Dependency B
    - PromiseKit

In this case, PromiseKit cannot be added to all three targets. I can get around adding it to the App but I am unable to get around it with Dependencies A and B.

RomanPodymov:
@vpolsinelli Do you have this problem only with PromiseKit?

niorko:
I would propose to make to products: `PromiseKit` (static) and `PromiseKitDyn` (dynamic) as multilayer architectures are quite common and you don't want to force anyone to use one or the other. Without this adjustment, we can't use this library either.

vpolsinelli:
I will try the workaround above and see if it works in my case. 

Updating the Package file to include two products is also a possible solution.

vpolsinelli:
I was able to work around this and no longer need this. Thanks for all the feedback.

PR #1169, opened 2020-10-04
Fix NSError().isCancelled

mxcl:


atom-wintermute:
@mxcl hey!
Could you please merge these changes and make new version for library?
we were faced with same crash recently.

Issue #1168, opened 2020-10-02
Xcode 12 - Warning related to iOS 8 support (SPM)


kikeenrique:
[PLEASE READ THE TROUBLESHOOTING GUIDE](https://github.com/mxcl/PromiseKit/blob/master/Documentation/Troubleshooting.md).

---

You read the guide but it didn’t help? OK, we’re here to help.

* Please specify the PromiseKit major version you are using
**PromiseKit 6.11.0**

* [Please format your code in triple backticks and ensure readable indentation](https://help.github.com/articles/creating-and-highlighting-code-blocks/)

* Please specify how you installed PromiseKit, ie. Carthage, CocoaPods, SwiftPM or other. If other provide DETAILED information about how you are integrating PromiseKit.
**SwiftPM**

If you ignore this template we will close your ticket and link to this template until you provide this necessary information. We cannot help you without it.

---

Using PromiseKit in SPM on Xcode 12 produces this warning:

_The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.0.99._

According to Xcode 12 release notes, iOS 8 is deprecated, check in https://developer.apple.com/documentation/xcode-release-notes/xcode-12-release-notes:

> Xcode 12 includes SDKs for iOS 14, iPadOS 14, tvOS 14, watchOS 7, and macOS Catalina 10.15.6. The Xcode 12 release supports on-device debugging for **iOS 9 and later**, tvOS 9 and later, and watchOS 2 and later. Xcode 12 requires an Intel-based Mac running macOS Catalina 10.15.4 or later.

The problem could be fixed changing the platforms supported in Package.swift
```
-        .macOS(.v10_10), .iOS(.v8), .watchOS(.v2), .tvOS(.v9)
+        .macOS(.v10_10), .iOS(.v9), .watchOS(.v2), .tvOS(.v9)
```

mxcl:
This is not a PromiseKit issue, I think more likely your Xcode project is misconfigured.

kikeenrique:
It's not a misconfiguration problem, it's a deprecation on Xcode 12 that conflicts with current Package.swift platform definition.
You just need to create a new project on Xcode 12, add SPM dependency on PromiseKit and you'll see the warning.

timbms:
I have the same issue as @kikeenrique. Would love to have it fixed.

gajjartejas:
[...]

PR #1170, opened 2020-10-05
Set iOS 9 as a minimal deployment target

serejahh:
Hey

Xcode 12 doesn't support iOS 8, so there're some annoying warnings:
image

What do you think about setting iOS 9 as a min version?

### DIFF
--- a/Package@swift-5.3.swift
+++ b/Package@swift-5.3.swift
@@ -1,0 +1,34 @@
+// swift-tools-version:5.3
+
+import PackageDescription
+
+let pkg = Package(name: "PromiseKit")
+pkg.platforms = [
+   .macOS(.v10_10), .iOS(.v9), .tvOS(.v9), .watchOS(.v2)
+]
+pkg.products = [
+    .library(name: "PromiseKit", targets: ["PromiseKit"]),
+]
+
+let pmk: Target = .target(name: "PromiseKit")
+pmk.path = "Sources"
+pmk.exclude = [
+    "AnyPromise.swift",
+    "AnyPromise.m",
+    "PMKCallVariadicBlock.m",
+    "dispatch_promise.m",
+    "join.m",
+    "when.m",
+    "NSMethodSignatureForBlock.m",
+    "after.m",
+    "hang.m",
+    "race.m",
+    "Deprecations.swift",
+    "Info.plist"
+]
+pkg.swiftLanguageVersions = [.v4, .v4_2, .v5]
+pkg.targets = [
+    pmk,
+    .testTarget(name: "A+", dependencies: ["PromiseKit"]),
+    .testTarget(name: "CorePromise", dependencies: ["PromiseKit"], path: "Tests/CorePromise"),
+]


kikeenrique:
I already reported this problem in #1168 but was closed as misconfiguration.
As it's not a configuration problem, would it be posible a reevaluation of the problem and get this PR merged?

gajjartejas:
When will be merged this?

mxcl:
What package manager are we talking about here? What version of PromiseKit?

mxcl:
The proposed change is breaking for Xcode < 12. We would have to make this change so it can only effect Xcode 12. Not sure if that is possible.

kikeenrique:
Do you mean it breaks compilation or it breaks support?
I suggest adding a new Package@swift.5.3.swift updating minimum iOS version to 9.0
That would only affect Xcode 12. 
I tested it on a new and clean project in Xcode 11 and on a new and clean project in Xcode 12.
What do you think?

mxcl:
> Do you mean it breaks compilation or it breaks support?

We support iOS 8, anyone who still has an app that uses PromiseKit and iOS 8 will update and find their project no longer builds which is a violation of our commitment to the versions of platforms will declare we support and thus we cannot remove that support without breaking our promise.

> I suggest adding a new Package@swift.5.3.swift updating minimum iOS version to 9.0

This is probably the only solution, though it's a solution via coincidence. I don’t think there is currently anything this will break though. If it works. I’m not sure `@5.3` is actually respected, usually they only allow these for tools-version bumps.

kikeenrique:
> > I suggest adding a new [Package@swift.5.3.swift](mailto:Package@swift.5.3.swift) updating minimum iOS version to 9.0
> 
> This is probably the only solution, though it's a solution via coincidence. I don’t think there is currently anything this will break though. If it works. I’m not sure `@5.3` is actually respected, usually they only allow these for tools-version bumps.

I've been looking for documentation for 2 hours about the Package Manifest rules for **@ operator** on the file, and manage to find something (also found out that you committed in [https://github.com/apple/swift-package-manager](swift-package-manager) too).
After my tests and reading Swift Package Manager [Handling Version-specific Logic](https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#handling-version-specific-logic), my understanding is that it should work. 

Another solution would be to silence the warning, but at the moment I don't know how to do that.

garyhooper:
Many Pods are exhibiting this same issue. Some are creating new releases for 9.0 and up, some are not in order to maintain 8.0 compatibility.

To silence the warnings, you can adjust your Podfile to change the deployment target.
E.g. see [this solution](https://github.com/facebook/react-native/issues/29255). It adjusts older versions to your current minimum supported version, in this case 9.0.
```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      if Gem::Version.new('9.0') > Gem::Version.new(config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'])
        config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '9.0'
      end
    end
  end
end
```

kikeenrique:
That solution is specific for CocoaPods, it's not applicable to SPM (Swift Package Manager).

kikeenrique:
As there has been no more proposals, @serejahh would you modify your PR with your proposed change moved to a new Package@swift.5.3.swift file?
With that committed, maybe @mxcl would reevaluate if this would be mergeable or not.

Regards!

Package@swift-5.3.swift
```diff
@@ -0,0 +1,34 @@
...
+    "PMKCallVariadicBlock.m",
+    "dispatch_promise.m",
+    "join.m",
+    "when.m",
+    "NSMethodSignatureForBlock.m",
+    "after.m",
+    "hang.m",
+    "race.m",
+    "Deprecations.swift",
+    "Info.plist"
```
serejahh, line 27: `Info.plist` has been added to fix the warning:
image

serejahh:
@kikeenrique I've added a package for Swift 5.3. Please check if it works ok for you

kikeenrique:
Just tested it, and it's working for me.

Package@swift-5.3.swift
kikeenrique: I guess this warning was due to the swift-tools-version bump from 5.0 to 5.3.

mohamede1945:
Just removing
```
pkg.platforms = [
   .macOS(.v10_10), .iOS(.v8), .tvOS(.v9), .watchOS(.v2)
]
```
should fix the problem and won't affect people with old versions of Xcode and/or supporting old platform versions

kikeenrique:
Is this PR mergeable? Could it be rechecked in CI?
The CI failed in a _pod_ step, and this code is unrelated to any _pod_ command check.

Kaspik:
@mxcl Any update on this? Warnings on XCode 12 after 4 months since release I kind of bothering... Thanks!

stuffmc:
Hey... The year is 2021... I think it would be okay to only support Xcode 12 and iOS > 8 like https://github.com/wtw-software/UTMConversion/issues/12 did. Thanks.

stuffmc:
Thanks for that! Now let's see what surprises comes from Xcode 12.6 or Xcode 13 ;-) Cheers!

PR #1171, opened 2020-10-06
Fix possible ambiguous use of when(fulfilled:) in tests

RomanPodymov:
Hello.
Thank you for PromiseKit.
I found out that in `testUnhandledErrorHandlerDoesNotFireForStragglers` and `testAllSealedRejectedFirstOneRejects` it is hard to determine which version of `when(fulfilled:)` is being used. As for me, there should be the **ambiguous use of when(fulfilled:)** error. My suggestion is to store `when(fulfilled: p1, p2, p3)`'s result in a variable, it helps to know which version of `when(fulfilled:)` is being used.


Tests/CorePromise/WhenTests.swift
```diff
@@ -218,7 +218,8 @@ class WhenTests: XCTestCase {
         let p2 = after(.milliseconds(100)).done { throw Error.straggler }
         let p3 = after(.milliseconds(200)).done { throw Error.straggler }
 
-        when(fulfilled: p1, p2, p3).catch { error -> Void in
+        let whenFulfilledP1P2P3: Promise<(Void, Void, Void)> = when(fulfilled: p1, p2, p3)
```
RomanPodymov, line 221: There can be also `let whenFulfilledP1P2P3: Promise<Void> = when(fulfilled: p1, p2, p3)`.

Tests/CorePromise/WhenTests.swift
```diff
@@ -241,7 +242,8 @@ class WhenTests: XCTestCase {
         let p2 = Promise<Void>(error: Error.test2)
         let p3 = Promise<Void>(error: Error.test3)
 
-        when(fulfilled: p1, p2, p3).catch { error in
+        let whenFulfilledP1P2P3: Promise<Void> = when(fulfilled: p1, p2, p3)
```
RomanPodymov, line 245: There can be also `let whenFulfilledP1P2P3: Promise<(Void, Void, Void)> = when(fulfilled: p1, p2, p3)`.

mxcl:
Forgive me, but I don’t see how this is ambiguous, ambiguous with which other version of when?

RomanPodymov:
Hello @mxcl 
For instance, in
```
let p1 = Promise<Void>(error: Error.test)
let p2 = after(.milliseconds(100)).done { throw Error.straggler }
let p3 = after(.milliseconds(200)).done { throw Error.straggler }

when(fulfilled: p1, p2, p3).catch { error -> Void in
    XCTAssertTrue(Error.test == error as? Error)
    ex1.fulfill()
}
```
As I can see in the debugger `func when<U: Thenable, V: Thenable, W: Thenable>(fulfilled pu: U, _ pv: V, _ pw: W) -> Promise<(U.T, V.T, W.T)>` is using here. But `public func when<U: Thenable>(fulfilled promises: U...) -> Promise<Void> where U.T == Void` is suitable here as well:
```
(when(fulfilled: p1, p2, p3) as Promise<Void>).catch { error -> Void in
    XCTAssertTrue(Error.test == error as? Error)
    ex1.fulfill()
}
```
Maybe there is a rule why `func when<U: Thenable, V: Thenable, W: Thenable>(fulfilled pu: U, _ pv: V, _ pw: W) -> Promise<(U.T, V.T, W.T)>` has a higher priority.

mxcl:
Oh ok. I’m not sure it matters per se, but why not.

PR #1186, opened 2021-01-24
Change minimum deployment target for SPM

denandreychuk:


mxcl:
Can you justify this, I believe we have to set v10 for one of our extensions.

PR #1190, opened 2021-02-25
doc: update documentation for doubling up Promises

centsent:
In the `Appendix` section, document about `Doubling up Promises` is outdated, a few developers follow the documentation (someone like me) would get stumped,  because they don't know how to wrap a new error while error occured in a Promise, so I update the documentation for it:

- `always` function was deprecated, should use `recover` instead here.
-  could wrap a new error with stack in `recover` function.


mxcl:
`recover` is wrong, `always` became `ensure`.

Either way, I don't understand what you are suggesting here.

mxcl:
Reopen if you have comment

PR #1193, opened 2021-03-17
ObjC AnyPromise and SwiftPackageManager (SPM)

DG0BAB:
This PR makes `AnyPromise` available to Objective-C and Swift code when used with SPM.

To achieve this, the PromiseKit package now consists of two targets. PromiseKit and PromiseKitObjC. The existence of PromiseKitObjC is almost irrelevant to the client because a simple `#import PromiseKit.h` can be used in Objective-C files to get access to `AnyPromise`. If the client wants to use modules, `@import PromiseKitObjC;` must be used. 

`AnyPromise` is now defined in Swift and uses the `@objc` attribute. It's made accessible to Objective-C, by adding the category `@interface AnyPromise (ObjC)` to it. All ObjC functionality defined in this category, delegates to the Swift `AnyPromise`. 

Additional changes to other files were necessary to make everything work in the PromiseKit Xcode Project and while build as a SwiftPackage. All tests succeeded in the Xcode Project.

### DIFF
--- a/Sources/AnyPromise.h
+++ b/Sources/AnyPromise.h
@@ -2,15 +2,13 @@
 #import <dispatch/dispatch.h>
 #import <PromiseKit/fwd.h>
 
-/// INTERNAL DO NOT USE
-@class __AnyPromise;
-
 /// Provided to simplify some usage sites
 typedef void (^PMKResolver)(id __nullable) NS_REFINED_FOR_SWIFT;
 
 
 /// An Objective-C implementation of the promise pattern.
-@interface AnyPromise: NSObject
+@interface AnyPromise (ObjC)
+
 
 /**
  Create a new promise that resolves with the provided block.
@@ -33,9 +31,6 @@ typedef void (^PMKResolver)(id __nullable) NS_REFINED_FOR_SWIFT;
  */
 + (instancetype __nonnull)promiseWithResolverBlock:(void (^ __nonnull)(__nonnull PMKResolver))resolveBlock NS_REFINED_FOR_SWIFT;
 
-
-/// INTERNAL DO NOT USE
-- (instancetype __nonnull)initWith__D:(__AnyPromise * __nonnull)d;
 
 /**
  Creates a resolved promise.
--- a/Sources/AnyPromise.h
+++ b/Sources/AnyPromise.h
@@ -1,6 +1,6 @@
 #import <Foundation/Foundation.h>
 #import <dispatch/dispatch.h>
-#import <PromiseKit/fwd.h>
+#import <fwd.h>
 
 // Different ways to make AnyPromise available
 //
--- a/Sources/AnyPromise.h
+++ b/Sources/AnyPromise.h
@@ -2,6 +2,27 @@
 #import <dispatch/dispatch.h>
 #import <PromiseKit/fwd.h>
 
+// Different ways to make AnyPromise available
+//
+// Import the Swift-Module if it was created already
+//
+// If building a Swift-Package just import PromiseKit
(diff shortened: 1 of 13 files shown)


DG0BAB:
As expected, the tests and/or build on Linux failed.

DG0BAB:
My mistake: Not the Linux-Tests are failing. It's the Pretest and I got irritated by the `--generate-linuxmain` option....

DG0BAB:
I tested it with Xcode 10.3 and got the same results as seen here while building with Travis. I currently don't know how to fix this and I don't have the time to further investigate this issue. However it works fine with Xcode 12. So maybe this PR could be used for some a later or special SPM version of PromiseKit. I don't know. Any help for fixing this Issue is greatly appreciated.

mxcl:
Rebase for new CI infrastructure

mxcl:
This is very intrusive. I cannot merge.

DG0BAB:
Yeah, a lot of changes were necessary to make it work. I’m already using my fork in production code for two of my projects and it works until any problems so far. 

If you are interested in merging let me know how to help.

Tests/Bridging/Infrastructure.swift
```diff
@@ -1,3 +1,4 @@
+import Foundation
```
mxcl, line 1: Presuambly unnecessary.

Sources/fwd.h
```diff
@@ -2,6 +2,7 @@
 #import <dispatch/dispatch.h>
 
 @class AnyPromise;
+
```
mxcl, line 5: Unnecessary.

mxcl:
We need to run the full CI against this. I'll make that possible with a comment or something.

Sources/AnyPromise.h
```diff
@@ -2,15 +2,35 @@
...
+// Import the Swift-Module if it was created already
+//
+// If building a Swift-Package just import PromiseKit
+//
+// In all other cases AnyPromise is still just a foward
+// declaration and needs to be defined. This is just to
+// silence a compile error which would avoid building
+// the Swift-Module on first builds.
+
+#if __has_include("PromiseKit-Swift.h")
```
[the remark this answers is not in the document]
DG0BAB: The Problem here was, that AnyPromise was not always defined at this point. But it must be defined here, because otherwise I could not declare a category on it. The include path is always different depending on what is currently build. Tests, Package, Xcode Project, etc. So I figured out that this seemed to work. I'm not very satisfied with it but finally it worked. Would appreciate a better solution. 

`self->d` isn't used anymore.  The idea I was following was defining `AnyPromise` in Swift and declaring a category in ObjC on that and delegate from the category to the Swift implementation.

Sources/fwd.h
DG0BAB: The newline or the forward declaration?

Tests/Bridging/Infrastructure.swift
DG0BAB: I'm sure there was a compiler error but I'll check again....

mrcararia:
Hi, 

is there any update about this issue? 
How is possibile to use AnyPromise in an ObjC class using the SPM version?

Thanks.

mxcl:
> is there any update about this issue?

This PR is too much. I have to maintain the results and this is beyond my ability to comprehend it without significant work. So I can't merge it.

> How is possibile to use AnyPromise in an ObjC class using the SPM version?

It is not possible without a patch like this. You could use the fork of OP. SwiftPM is decentralized by design.

PR #1199, opened 2021-04-30
Removed Swift Compiler Warnings

johnarnaou:
- Using AnyObject instead of class in Thenable protocol
- Changed root variable to a let


rgoldberg:
This seems to have already been committed & merged in 67e588c804a4c8c8b902cb13625bb476134c72d5.

mxcl:
Merged separately with 6.14.0

PR #1200, opened 2021-04-30
Removed Target Integrity warnings

johnarnaou:
- Deployment target changed from iOS 8.0 to iOS 9.0


lastthyme (approved):

mxcl:
Cannot without a major version bump since it breaks all our extensions (with Carthage), so we'll need to coordinate a bump to all of those simultaneously.

PR #1211, opened 2021-05-25
Fix [[AnyPromise alloc] init]

### LINKED ISSUE #1175: Make it illegal to construct AnyPromise() in Swift without arguments
* PromiseKit 6.13.1, installed via CocoaPods

I recently ran into a nasty crashing bug in my codebase because I had a `return AnyPromise()` in some edge case. I wasn't aware that it was illegal to construct `AnyPromise()` without any arguments.

`AnyPromise`'s default argument-less `init` initializer is correctly marked as `unavailable` in the Objective-C header (see https://github.com/mxcl/PromiseKit/blob/master/Sources/AnyPromise.h#L295).
When trying to call it from Objective-C with `[[AnyPromise alloc] init]` the compiler correctly complains. However, there's no warning and no compiler error when illegally constructing it from Swift using `AnyPromise()`. Is there a way to work around for this? I've tried a few, but to no avail. Shouldn't the Swift compiler pick up the Objective-C `unavailable` declaration in `AnyPromise.h`? So is there any solution we could implement for this? 

Here's a related discussion with an Apple Engineer: https://developer.apple.com/forums/thread/650054?login=true&page=1#648364022

RomanPodymov:
Hello.
Thank you for PromiseKit.
Fix #1175. It looks like the attributes from Objective-C categories are not visible in Swift.
Found related bug [here](https://bugs.swift.org/browse/SR-7641).

### DIFF
--- a/Sources/AnyPromise.h
+++ b/Sources/AnyPromise.h
@@ -213,6 +213,17 @@ typedef void (^PMKResolver)(id __nullable) NS_REFINED_FOR_SWIFT;
 */
 - (instancetype __nonnull)initWithResolver:(PMKResolver __strong __nonnull * __nonnull)resolver NS_REFINED_FOR_SWIFT;
 
+/**
+ Unavailable methods
+ */
+
+- (instancetype __nonnull)init __attribute__((unavailable("It is illegal to create an unresolvable promise.")));
++ (instancetype __nonnull)new __attribute__((unavailable("It is illegal to create an unresolvable promise.")));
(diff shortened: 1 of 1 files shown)


mxcl:
Merged since macOS passed. The failure was my error, fixed in master for new PRs. Thanks for the PR 🙏🏻

jfahrenkrug:
@RomanPodymov @mxcl Fantastic, thanks so much!

PR #1215, opened 2021-05-26
Fixed isCancelled crash

dtaylor1701:
`Swift.Error` was not being bridged properly to `NSError`, resulting in a selector not found exception when "domain" and "code" values were accessed via a cast only to `AnyObject`. This casts as `NSError` so that the bridging provided by the compiler works as expected.

I wasn't able to determine if this is a change in the recent Swift release or some other more subtle change, but this approach is clear and works well in my testing.

I was able to reproduce the crash this fixes with a simple playground containing:
``` swift 
import Foundation

class TestingError: Error { }

let error = TestingError()

let domain = (error as? AnyObject)?.value(forKey: "domain") as? String
```

### DIFF
--- a/Sources/Error.swift
+++ b/Sources/Error.swift
@@ -89,8 +89,9 @@ extension Error {
             return true
         } catch {
         #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
-            let domain = (error as AnyObject).value(forKey: "domain") as? String
-            let code = (error as AnyObject).value(forKey: "code") as? Int
+            let error = error as NSError
+            let domain = error.domain
+            let code = error.code
             return ("SKErrorDomain", 2) == (domain, code)
         #else
             return false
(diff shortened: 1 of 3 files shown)


Sources/Error.swift
```diff
@@ -87,14 +87,16 @@ extension Error {
             return true
         } catch CocoaError.userCancelled {
             return true
+        } catch let error as NSError {
+            #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
```
dtaylor1701, line 91: I'm not entirely sure this check is necessary anymore.

dtaylor1701:
Huh, the macOS tests all pass when running locally.

mxcl:
> Huh, the macOS tests all pass when running locally.

The progress test fails in CI sometimes. I need to figure out why.

dtaylor1701:
Not sure what's going on with that codecov/project check.

PR #1265, opened 2021-09-21
note that cancel doesn't trigger catch block

ConfusedVorlon:


ConfusedVorlon:
sorry - realised I was targeting v7

ConfusedVorlon:
6.15.3 is the same...

mxcl:
This just repeats the same text from two paragraphs above.

ConfusedVorlon:
I guess so - I was just trying emphasise it.

mxcl:
I think emphasizing it is a good idea. So stick the `**important**` with the other paragraph?

PR #1266, opened 2021-09-21
note cancellation error as possible reason for promise never resolving

ConfusedVorlon:


ConfusedVorlon:
sorry - realised I was targeting v7

ConfusedVorlon:
6.15.3 is the same...

PR #1269, opened 2021-09-23
func race(fulfilled:) for Objective-C

RomanPodymov:
Hello.
Thank you for PromiseKit.
Add `func race(fulfilled:)` for Objective-C.

### DIFF
--- a/Sources/race.swift
+++ b/Sources/race.swift
@@ -91,6 +91,7 @@ public func race<U: Thenable>(fulfilled thenables: [U]) -> Promise<U.T> {
                     }
                 case .fulfilled(let value):
                     guard rp.isPending else { return }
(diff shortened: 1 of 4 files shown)


Sources/race.swift
```diff
@@ -91,7 +91,6 @@ public func race<U: Thenable>(fulfilled thenables: [U]) -> Promise<U.T> {
                     }
                 case .fulfilled(let value):
                     guard rp.isPending else { return }
-                    countdown = 0
```
RomanPodymov, line 94: An unnecessary line.

mxcl: Probably, but there is no reason to remove it. It is possible this is required. Please leave it.

PR #1280, opened 2021-12-08
possible cocoa pods sources fix

apps4everyone:
possible cocoa pods sources fix
maybe you can run this on CI?


mxcl:
I added CI for the `.podspec` file so if you rebase it *should* run.

PR #1286, opened 2022-02-19
ensureInBackground + QOS_CLASS_UNSPECIFIED

RomanPodymov:
Hello.
Thank you for PromiseKit.
Add `ensureInBackground`. Use `QOS_CLASS_UNSPECIFIED` instead of 0 where applicable. Methods ordering `something` -> `somethingOn` -> `somethingInBackground`.


mxcl:
The patch has a number of unrelated movements in the header, this makes it hard for me to be sure that nothing has been changed that is incorrect. Can we undo these parts of the diff?

mxcl (approved):

PR #1288, opened 2022-03-25
prevent dependency cycles by creating headers before compiling sources

michaelknoch:
Creating the headers before compiling the sources prevents potential dependency cycles. 

This started happening for us after updating xcode to version 13.3. I double checked this and there obviously is no dependency from PromiseKit to our libraries, so the error doesn't make much sense to me. Changing the order of the build phases reliably fixes the problem.

Screenshot 2022-03-25 at 16 47 06

### DIFF
--- a/PromiseKit.xcodeproj/project.pbxproj
+++ b/PromiseKit.xcodeproj/project.pbxproj
@@ -586,9 +586,9 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = 63B0AC5F1D595E1B00FA21D9 /* Build configuration list for PBXNativeTarget "PromiseKit" */;
 			buildPhases = (
+				63B0AC541D595E1B00FA21D9 /* Headers */,
 				63B0AC521D595E1B00FA21D9 /* Sources */,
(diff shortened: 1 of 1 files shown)


mxcl:
I seem to remember that there is a reason for the order (for the objc portion which contains trickery).

Naturally you can’t comment this sort of thing so it's lost in commit history.

However, if it's generally causing breakage I'm not sure what else we can do but merge it.

michaelknoch:
I noticed that this is also now the default order of build phases when creating a new framework via xcode. We use a few thin layers of objective c in our apps and so far I was not running into issues.

Screenshot 2022-04-01 at 12 59 39

symorton:
We are also experiencing this after the update. Moving the Header locally fixes the issue but we would love to not have to worry about it.

mxcl:
Why didn’t any CI run? ugh

michaelknoch:
no idea, maybe CI doesn't run for PRs from forks? Feel free to recreate a new PR @mxcl

michaelknoch:
thanks @mxcl

PR #1291, opened 2022-07-14
Async/await

RomanPodymov:
Hello.
Thank you for PromiseKit.
Added compatibility with async/await.


mxcl (approved):

mxcl (changes requested):
Needs to pass CI, our strategy with these typically is to add the `#if` inside the test bodies so that they pass because they have no content.

RomanPodymov:
> Needs to pass CI, our strategy with these typically is to add the `#if` inside the test bodies so that they pass because they have no content.

Fixed.

mxcl (approved):

mxcl:
v6.18.0 released. Thanks for the contribution.

mxcl:
K possibly this is just we now require a newer Xcode https://stackoverflow.com/questions/69284960/ios-concurrency-is-only-available-in-ios-15-0-0-or-newer-in-protocol

PR #1293, opened 2022-08-15
Fix #1292

### LINKED ISSUE #1292: error: concurrency is only available in iOS 15.0.0 or newer    func async() async throws -> T {
error: concurrency is only available in iOS 15.0.0 or newer
     [exec]      [exec]     func async() async throws -> T {

RomanPodymov:
Hello.
Thank you for PromiseKit.
Tried to fix #1292.

### DIFF
--- a/Sources/Async.swift
+++ b/Sources/Async.swift
@@ -1,4 +1,4 @@
-#if swift(>=5.5)
+#if swift(>=5.5) && canImport(_Concurrency)
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
 public extension Guarantee {
     func async() async -> T {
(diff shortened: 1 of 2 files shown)


mxcl:
Looks great!

I wish we could test for this, sadly I don’t think it's trivial.

mxcl:
Looks like the `#if(_concurrency) needs to be inside the compiler check`: https://github.com/mxcl/PromiseKit/runs/7840727401?check_suite_focus=true

RomanPodymov:
> Looks like the `#if(_concurrency) needs to be inside the compiler check`: https://github.com/mxcl/PromiseKit/runs/7840727401?check_suite_focus=true

Fixed.

mxcl:
6.18.1 released. Thanks again.

PR #1311, opened 2023-03-31
Update to Xcode 14 / Swift 5.7

### LINKED ISSUE #1308: Problems building with Xcode 14.3 beta2
I got this error when building with Carthage and Xcode 14.3 beta2 

```
/Users/jose/Documents/3.7.9/ice/Carthage/Checkouts/PromiseKit/PromiseKit.xcodeproj: warning: The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.10, but the range of supported deployment target versions is 10.13 to 13.3.99. (in target 'PromiseKit' from project 'PromiseKit')
** ARCHIVE FAILED **
```

threema-jarem:
This requires dropping support for iOS/tvOS <11, macOS < 10.13, watchOS < 4.

Resolve #1308


mxcl:
Yeah I don't think changing the podspec is necessary. CocoaPods generates a compliant xcodeproj. By changing the pod spec you are potentially breaking anyone who uses this framework with old Xcodes to deliberately target old iOS versions.

The xcodeproj changes are ok though as there is no choice with Carthage.

mxcl (approved):

mxcl:
Deploying 6.22.1.

Which is really an incorrect semver bump since this changes platform requirements. But it's only for Carthage so…

Issue #1317, opened 2023-05-31
Cocoapods's pods that depend on PromiseKit 6.22.1+ fail to `pod lib lint` while using Xcode 14.3


crsantos:
Version: 6.22.1
Installed via: Cocoapods

Hi,

I'm following up after this issue https://github.com/mxcl/PromiseKit/issues/1308 and the PR that fixed it: https://github.com/mxcl/PromiseKit/pull/1311.

That PR fixed Xcode 14.3 builds, but while using Cocoapods, as you said on the issue, it creates a temporary project for the linting process, based on the `podspec`. 

Any pod that depends on PromiseKit will fail on linting due to the iOS 8.0 deployment target.

I know you referred here that dropping support for older versions is tricky, https://github.com/mxcl/PromiseKit/pull/1311#issuecomment-1491892671

But this also blocks people from using Xcode 14.3 and beyond, hence my question:

are we good to also bump the [podspec's deployment target](https://github.com/mxcl/PromiseKit/blob/v6/PromiseKit.podspec#L27) to `11.0`? 

(Same as https://github.com/mxcl/PromiseKit/pull/1311 did for the xcodeproj)

### Log

This can be replicated with a simple `pod lib lint` on the repo:

<details>
  <summary>Expand</summary>

```shell
  PromiseKit on  v6 via 🏎
➜  pod lib lint

 -> PromiseKit (6.22.1)
    - NOTE  | url: The URL (https://twitter.com/mxcl) is not reachable.
    - ERROR | [PromiseKit/Accounts, PromiseKit/Alamofire, PromiseKit/AddressBook, and more...] xcodebuild: Returned an unsuccessful exit code. You can use `--verbose` for more information.
    - NOTE  | [PromiseKit/Accounts, PromiseKit/Alamofire, PromiseKit/AddressBook, and more...] xcodebuild:  note: Using codesigning identity override: -
    - NOTE  | [PromiseKit/Accounts, PromiseKit/Alamofire, PromiseKit/AddressBook, and more...] xcodebuild:  note: Building targets in dependency order
    - NOTE  | [PromiseKit/Accounts, PromiseKit/Alamofire, PromiseKit/AddressBook, and more...] xcodebuild:  clang: error: linker command failed with exit code 1 (use -v to see invocation)
[...]

PR #1318, opened 2023-05-31
Bumping iOS deployment target to 11.0 (10.0 after discussion)

crsantos:
## What changes

* Bumping iOS deployment target to `11.0`
* Also osx to `10.13`, tvos to `11.0` and watchos to `4.0`
* Fixed some warnings due to the bump
* Bumped version to `6.23.0` (confirm me if I should do it or not, I can revert if needed)

## Why do we need this PR

This PR fixes the issue reported on https://github.com/mxcl/PromiseKit/issues/1317, while using Xcode `14.3` to validate the podspec.

Every pod that depends on `PromiseKit/CorePromise` will fail to lint due to what is being explained here https://github.com/mxcl/PromiseKit/issues/1308#issuecomment-1490190549.

**TL;DR**
_Apple has stopped shipping libarclite_iphoneos.a separately [1](https://developer.apple.com/forums/thread/725300)_

As https://github.com/mxcl/PromiseKit/pull/1311 only fixes the xcodeproj, this PR fixes the podspec.

### Implications

With this change it means that consumers need a minimum deployment target of `11.0`, but this will enable future Xcode versions since `14.3`.

### DIFF
--- a/PromiseKit.podspec
+++ b/PromiseKit.podspec
@@ -24,10 +24,10 @@ Pod::Spec.new do |s|
   # CocoaPods requires us to specify the root deployment targets
   # even though for us it is nonsense. Our root spec has no
   # sources.
-  s.ios.deployment_target = '8.0'
-  s.osx.deployment_target = '10.10'
-  s.watchos.deployment_target = '2.0'
-  s.tvos.deployment_target = '9.0'
+  s.ios.deployment_target = '10.0'
+  s.osx.deployment_target = '10.13'
+  s.watchos.deployment_target = '4.0'
+  s.tvos.deployment_target = '10.0'
 
   s.pod_target_xcconfig = {
     'OTHER_SWIFT_FLAGS' => '-DPMKCocoaPods',
@@ -38,19 +38,8 @@ Pod::Spec.new do |s|
     ss.exclude_files = 'Extensions/Accounts/Sources/*.plist'
     ss.ios.frameworks = ss.osx.frameworks = 'Accounts'
     ss.dependency 'PromiseKit/CorePromise'
-    ss.ios.deployment_target = '8.0'
-    ss.osx.deployment_target = '10.10'
-  end
-
-  s.subspec 'Alamofire' do |ss|
-    ss.source_files = 'Extensions/Alamofire/Sources/**/*'
-    ss.exclude_files = 'Extensions/Alamofire/Sources/*.plist'
-    ss.dependency 'Alamofire', '~> 4.0'
-    ss.dependency 'PromiseKit/CorePromise'
-    ss.ios.deployment_target = '8.0'
-    ss.osx.deployment_target = '10.11'
-    ss.watchos.deployment_target = '2.0'
-    ss.tvos.deployment_target = '9.0'
+    ss.ios.deployment_target = '10.0'
+    ss.osx.deployment_target = '10.13'
   end
 
   s.subspec 'AddressBook' do |ss|
@@ -58,7 +47,7 @@ Pod::Spec.new do |s|
     ss.exclude_files = 'Extensions/AddressBook/Sources/*.plist'
     ss.ios.frameworks = 'AddressBook'
     ss.dependency 'PromiseKit/CorePromise'
-    ss.ios.deployment_target = '8.0'
+    ss.ios.deployment_target = '10.0'
   end
 
   s.subspec 'AssetsLibrary' do |ss|
@@ -66,7 +55,7 @@ Pod::Spec.new do |s|
     ss.exclude_files = 'Extensions/AssetsLibrary/Sources/*.plist'
     ss.ios.frameworks = 'AssetsLibrary'
     ss.dependency 'PromiseKit/CorePromise'
-    ss.ios.deployment_target = '8.0'
+    ss.ios.deployment_target = '10.0'
   end
 
   s.subspec 'AVFoundation' do |ss|
@@ -74,18 +63,7 @@ Pod::Spec.new do |s|
     ss.exclude_files = 'Extensions/AVFoundation/Sources/*.plist'
     ss.ios.frameworks = 'AVFoundation'
     ss.dependency 'PromiseKit/CorePromise'
-    ss.ios.deployment_target = '8.0'
-  end
-
-  s.subspec 'Bolts' do |ss|
-    ss.source_files = 'Extensions/Bolts/Sources/**/*'
-    ss.exclude_files = 'Extensions/Bolts/Sources/*.plist'
-    ss.dependency 'PromiseKit/CorePromise'
-    ss.dependency 'Bolts', '~> 1.9.0'
-    ss.ios.deployment_target = '8.0'
-    ss.osx.deployment_target = '10.10'
-    ss.watchos.deployment_target = '2.0'
-    ss.tvos.deployment_target = '9.0'
+    ss.ios.deployment_target = '10.0'
   end
 
   s.subspec 'CloudKit' do |ss|
@@ -94,9 +72,9 @@ Pod::Spec.new do |s|
     ss.frameworks = 'CloudKit'
     ss.dependency 'PromiseKit/CorePromise'
     ss.ios.deployment_target = '10.0'
-    ss.osx.deployment_target = '10.12'
+    ss.osx.deployment_target = '10.13'
     ss.tvos.deployment_target = '10.0'
-    ss.watchos.deployment_target = '3.0'
+    ss.watchos.deployment_target = '4.0'
   end
 
   s.subspec 'CoreBluetooth' do |ss|
@@ -104,9 +82,9 @@ Pod::Spec.new do |s|
     ss.exclude_files = 'Extensions/CoreBluetooth/Sources/*.plist'
     ss.ios.frameworks = ss.osx.frameworks = ss.tvos.frameworks = 'CoreBluetooth'
     ss.dependency 'PromiseKit/CorePromise'
-    ss.ios.deployment_target = '8.0'
-    ss.osx.deployment_target = '10.10'
-    ss.tvos.deployment_target = '9.0'
+    ss.ios.deployment_target = '10.0'
+    ss.osx.deployment_target = '10.13'
+    ss.tvos.deployment_target = '10.0'
   end
 
   s.subspec 'CorePromise' do |ss|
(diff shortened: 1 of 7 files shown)


Tests/CorePromise/ThenableTests.swift
```diff
@@ -200,11 +200,13 @@ class ThenableTests: XCTestCase {
     }
 
     func testLastValueForEmpty() {
-        XCTAssertTrue(Promise.value([]).lastValue.isRejected)
+        let values = [String]()
```
crsantos, line 203: just forced a type

nunogoncalves (approved):

simaoseica (approved):

RomanPodymov (approved):

erickjung (approved):

fabioalmeida (approved):

mxcl (approved):

PromiseKit.podspec
```diff
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
   s.name = "PromiseKit"
 
-  s.version = '6.22.1'
+  s.version = '6.23.0'
```
mxcl, line 4: don't change this, we change it in CI/CD

mxcl:
Seems we're missing some deployment targets? https://github.com/mxcl/PromiseKit/actions/runs/5182487581/jobs/9339327669?pr=1318#step:3:22

crsantos:
> Seems we're missing some deployment targets? https://github.com/mxcl/PromiseKit/actions/runs/5182487581/jobs/9339327669?pr=1318#step:3:22

Oh,

Here's the issue as I'm seeing it, as I didn't know about these subspecs being submodules.

We have a lot of submodules from different libs/repos (e.g.: `PromiseKit/Alamofire`), but they're exposed as subspecs of `PromiseKit`,

As they all depend on a past version of `PromiseKit`, e.g.: [here](https://github.com/PromiseKit/PMKAlamofire/blob/master/Cartfile#L1)

```ruby
github "mxcl/PromiseKit" ~> 6.13.0
```

Maybe I'm seeing it wrong, but we would need a new version of `PromiseKit` on every single submodule, compiling on every Xcode, including 14.3 (using `6.22.1` would work), but they are 23 extension, AFAIK:

* AVFoundation
* Accounts
* AddressBook
* Alamofire
* AssetsLibrary
* Bolts
* CloudKit
* CoreBluetooth
* CoreLocation
* EventKit
* Foundation
* HealthKit
* HomeKit
* MapKit
* MessagesUI
* OMGHTTPURLRQ
* Photos
* QuartzCore
* Social
* StoreKit
* SystemConfiguration
* UIKit
* WatchConnectivity

So we need 23 PRs 😬

crsantos:
Hi, apologies for the delay.

> > But please, confirm if I'm getting this right.
> 
> no that only effects Carthage. The CocoaPod is entirely defined by the `PromiseKit.podspec` file.
> 
> We don't have the same problems with Carthage (which ofc backs up my original opinion that this is a CocoaPods issue that we are being forced to work around).


You're totally right, the dependencies of submodules/subspecs are being defined using Carthage, and it's a completely different subject.

But I have found the issue of the failed `pod lib lint`:

* As [PromiseKit/Alamofire](https://github.com/PromiseKit/PMKAlamofire) depends on [Alamofire](https://github.com/Alamofire/Alamofire) itself, specifically, v`4.0`, a very old version from 2019, [it targets a deployment target of iOS `8.0`](https://github.com/Alamofire/Alamofire/blob/4.9.1/Alamofire.podspec#L13).
* Even if we update [PromiseKit/Alamofire](https://github.com/PromiseKit/PMKAlamofire) to the latest major version of Alamofire (it's a major upgrade from `4.9.1` to `5.7.0`), it still requires iOS `10.0` ([apparently they reverted the bump to iOS `11.0`](https://github.com/Alamofire/Alamofire/commit/236844f551a72b7528b1634c9b8266b17f5abcb3)), and we can't have this spec passing the lint due to this.

What needs to be done, AFAIK:

1. If this is a cocoapods issue, having this fixed on cocoapods
1. If this is due to Xcode `14.3`, (because it works fine with `14.2`, and it's not an issue):
* Bump [Alamofire](https://github.com/Alamofire/Alamofire) deployment target to iOS `11.0` and launch that version
* Update [PromiseKit/Alamofire](https://github.com/PromiseKit/PMKAlamofire) to use the new major version that will support Xcode `14.3`
* Only after that, I can point this repo to the new version of [PromiseKit/Alamofire](https://github.com/PromiseKit/PMKAlamofire) that depends on the fixed version of [Alamofire](https://github.com/Alamofire/Alamofire).


Does my explanation make sense?

mxcl:
We can drop Alamofire and release PMK 7 (this existing PMK7 rewrite branch was never released and should probs be abandoned at this point).

mxcl:
The fact that Alamofire doesn't support iOS 11 makes me wonder why we cannot do the same? I know they are 10 and we are 8, but if they don't go to 11 why must we?

crsantos:
> The fact that Alamofire doesn't support iOS 11 makes me wonder why we cannot do the same? I know they are 10 and we are 8, but if they don't go to 11 why must we?

That may work too, using 10 as minimum, as they're saying here https://github.com/Alamofire/Alamofire/issues/3716#issuecomment-1546248048, so it may lint, with just a warning. 
I didn't know that was possible otherwise that would be my first try.

Let me validate that it lints with 10 ⌛ 

That can be one option, but it still require PMKAlamofire to be updated to use the last major of Alamofire 5.7.0.

But if you think we can drop PMKAlamofire subspec, we can go with the 7.0 branch. I'm just not sure what's in there, I'll have to check.

mxcl:
eek, not sure what else to do.

Basically, remove them and see if this passes (remove the PMK version change, our automation will handle that). If it passes I will merge and release a v7.

crsantos:
> eek, not sure what else to do.
> 
> Basically, remove them and see if this passes (remove the PMK version change, our automation will handle that). If it passes I will merge and release a v7.

Thanks for the feedback, Max 🙇 

So, bumping iOS deployment target to `10.0` and removing the 3 specs, it works on the lint, but of course, with `--allow-warnings` enabled.

```shell
pod lib lint --allow-warnings --fail-fast
```

I'll update the MR to conform with that and let's see what CI tells us 😃

mxcl:
v8 https://github.com/mxcl/PromiseKit/actions/runs/5313482075

mxcl:
failed due to old images, running again.

v8 because v7 was published and some people might depend on it.

crsantos:
> failed due to old images, running again.

If you need any action from my side, please shout 🙏

crsantos:
Thanks for the help @mxcl and @RomanPodymov 🙏

mxcl:
Apologies for this being long overdue

oklimberg:
Now that subspec for Alamofire has been removed with version 8.0.0, can I still use PromiseKit and Alamofire with CocoaPods, or do I need to switch to SPM?
What was the argument against updating the Alamofire subspec (using Alamofire 5.x)? SPM Package PMKAlamofire version 4.0.0 does support Alamofire 5.x.
I also didn't quite understand why the decision was still on iOS 10 for the deployment target, when Apple requires Xcode 14 for updates/submissions and this version only supports iOS 11 (https://developer.apple.com/news/?id=jd9wcyov)
So to get rid of warnings I still have to use a post install hook for CocoaPods.

crsantos:
Hi @oklimberg 

> Now that subspec for Alamofire has been removed with version 8.0.0, can I still use PromiseKit and Alamofire with CocoaPods, or do I need to switch to SPM?

As of v8, as the podspec doesn't have that subspec, **no**, you can't use Cocoapods if you need the Alamofire subspec.
But you can target v6 or v7 for that.

> What was the argument against updating the Alamofire subspec (using Alamofire 5.x)? 

I think we didn't discuss in detail the why, maybe we should, but I'll explain our reasoning below.

> SPM Package PMKAlamofire version 4.0.0 does support Alamofire 5.x.

I was not aware of that 🤔 [and you're right](https://github.com/PromiseKit/PMKAlamofire/blob/master/Package.swift#LL16C13-L16C13). But bear in mind that it's not 5.5.2 that fixes the issue with iOS 10/11 deployment target, it's just [5.7.1](https://github.com/Alamofire/Alamofire/releases/tag/5.7.1)

But, my 2 cents on why I didn't do this:

`PMKAlamofire` was locked to `Alamofire` `4.0`, [as you can see on the discussion above](https://github.com/mxcl/PromiseKit/pull/1318#issuecomment-1587238635), and it needs the adoption of that major version 5.7.2. Why do I say that? That it needs adoption of major:

* Did you test with `PMKAlamofire` with Xcode 14.3? 
  * because `PMAlamofire`  wasn't compiling when bumping that major version of AF, changes were required.

> I also didn't quite understand why the decision was still on iOS 10 for the deployment target, when Apple requires Xcode 14 for updates/submissions and this version only supports iOS 11 (https://developer.apple.com/news/?id=jd9wcyov)

The explanation is [on this comment](https://github.com/mxcl/PromiseKit/pull/1318#issuecomment-1592413586).
> So to get rid of warnings I still have to use a post install hook for CocoaPods.

IMHO, post install hook solves the issue for a `Podfile`, and it works.
But a spec not linting because a podspec dependency targets older deployment targets is something, is a real problem.

Nevertheless, the submodule is still there. 
If everyone agrees that it should stil be considered and maintained, it's just a question of contributing with a PR to update `PMKAlamofire` work with  AF `5.7.1` and then re-introduce the spec with proper minimum deployment targets.

oklimberg:
Thank you very much for your time explaining the reasoning.
I will have a look, if I am able to provide a PR for PromiseKit 8 and Alamofire 5 in general and if the linting issue will require
Alamofire 5.7.1

crsantos:
Maybe Alamofire 5 doesn’t need any change.
Only PMKAlamofire to adopt the changes AF requires for that v5.7.1. But that version is still iOS 10, they decided to keep the 10, reverted from 11.

In terms of the changes, it’s a matter of bumping AF dependency on PMKAlamofire’s Cartfile and compile to replicate the issue I had.

mxcl:
Bringing back the Alamofire subspec is welcome. PR should be pretty simple.

We could not keep Alamofire and change the deployment targets so it had to be removed or upgraded. We wanted to give people a working PromiseKit release ASAP so we removed it.

@oklimberg I can assist with the work if you need it.

crsantos:
@oklimberg I can also assist on this if you need

VaryHice:
Hello @mxcl and @crsantos!
I tried to bring back Alamofire subspec, but encountered compilation issue due to recent changes in PMKAlamore imports. I suppose they were done for a reason, probably they affect Carthage somehow - badly I'm not familiar enough. Would be glad if you could help and check my PR to PMKAlamofire: https://github.com/PromiseKit/PMKAlamofire/pull/52. After solving this issue I could make PR to PromiseKit bringing back Alamofire subspec.

PR #1348, opened 2024-09-18
Preparing for Swift 6

RomanPodymov:


mxcl:
Note these are breaking changes that require a major version bump *and* mean we only support Swift >=6

PR #1352, opened 2025-01-05
Android support

marcprux:
This PR adds support for Android, as well as a CI action to run the unit tests against an Android emulator.


mxcl:
thanks! ci for android fails

marcprux:
> ci for android fails

The failure is:

```
/home/runner/work/PromiseKit/PromiseKit/Tests/CorePromise/AfterTests.swift:40: error: AfterTests.testPositive : Asynchronous wait failed - Exceeded timeout of 2.0 seconds, with unfulfilled expectations: 
```

It passed in my fork's CI run, and I suspect that if you re-ran the workflow once or twice that it would work for you. I think the problem is just that the Android emulator is slow, so timing tests like these need a little more padding.

I could bump up the test tolerance (perhaps just for Android) in places like https://github.com/mxcl/PromiseKit/blob/master/Tests/CorePromise/AfterTests.swift#L40C29-L40C36 if you think that would be OK?

mxcl:
Oh ok, that's fine. I'll merge and fix the CI stuff separately.

